PR #81, opened 2019-02-06
Hernan604/feature trust browser

hernan604:
Displays a checkbox on login screen to trust browser when 2FA is enabled


petersalomonsen:
Just need to figure out why the 2fa login test does not pass @hernan604

PR #111, opened 2019-04-02
Cache folderlist when offline

petersalomonsen:
This makes the folder list available also when offline.


gtandersen:
Excellent @petersalomonsen!

PR #121, opened 2019-04-10
Add a Calendar app

tadzik:


src/styles.scss
```diff
@@ -1,5 +1,7 @@
 @import '../node_modules/@angular/material/theming';
 @import '../iconfont/material-icons.css';
+@import "../node_modules/angular-calendar/css/angular-calendar.css";
+@import "~ng-pick-datetime/assets/style/picker.min.css";
```
petersalomonsen: I've never seen imports starting with ~ but is that intended? @tadzik

tadzik: Looks like it's a feature: https://stackoverflow.com/questions/39535760/what-does-a-tilde-in-a-css-url-do/39535907#39535907

I guess it's best to either use ~ everywhere or switch ng-pick-datetime to "absolute" paths like in other lines.

PR #126, opened 2019-04-16
folderlist(fix): handle draft subfolders

petersalomonsen:
folder list crashed if there were draft subfolders.

This fix will not show any draft subfolders, but prevent the folderlist from crashing ( and showing no folders at all) if there are draft subfolders.

### DIFF
--- a/src/app/folder/folderlist.component.ts
+++ b/src/app/folder/folderlist.component.ts
@@ -81,7 +81,7 @@ export class FolderListComponent {
 
         const foldertreedataobservable = this.folders
             .pipe(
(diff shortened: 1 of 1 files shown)


gtandersen:
@petersalomonsen Great, but can we fetch the name of the Drafts folder from the `site.conf` config file?

tadzik:
A minor nitpick, now that I learned about it: I think the commit message is the wrong way around: should be fix(folderlist), as in https://github.com/runbox/runbox7/blob/master/CONTRIBUTING.md#commit-message-format :)

tadzik:
Updated and reopened at https://github.com/runbox/runbox7/pull/374

PR #134, opened 2019-04-29
Add contact groups to compose autocompletion

tadzik:


petersalomonsen:
@tadzik How do I create a contact group to test this?

tadzik:
@petersalomonsen when editing a contact you have an option to assign them to a group. You can also create a new one in that same dropdown menu.

src/app/compose/mailrecipientinput.component.ts
```diff
@@ -53,18 +54,24 @@ export class MailRecipientInputComponent implements OnInit, AfterViewInit {
...
                 if (searchtext) {
-                    const lowercaseSearchText = searchtext.toLowerCase();
+                    // I know that searchtext is supposed to be a string,
+                    // but somehow it ends up being the recipient object
+                    // just as you click the autocomplete suggestion. Thus, .toString()
+                    const lowercaseSearchText = searchtext.toString().toLowerCase();
                     this.filteredRecipients.next(
                         recipients.filter(recipient =>
-                            recipient.toLowerCase().indexOf(lowercaseSearchText) > -1
+                            recipient.name.toLowerCase().indexOf(lowercaseSearchText) > -1
```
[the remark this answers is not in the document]
tadzik: Hmm. Is there anything specific about the contact? Did it not happen before this change?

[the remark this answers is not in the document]
petersalomonsen: This was after adding a contact group. Not having this issue with the current production version (which does not support contact groups in compose autocompletion)

src/app/compose/mailrecipientinput.component.ts
```diff
@@ -53,18 +54,24 @@ export class MailRecipientInputComponent implements OnInit, AfterViewInit {
...
         recipientservice.recipients.subscribe((recipients) => {
 
         // Listen to search text input and popup suggestions from recipient list
         this.searchTextFormControl.valueChanges
             .pipe(debounceTime(50))
             .subscribe((searchtext: string) => {
                 if (searchtext) {
-                    const lowercaseSearchText = searchtext.toLowerCase();
+                    // I know that searchtext is supposed to be a string,
+                    // but somehow it ends up being the recipient object
```
petersalomonsen: Would we want to figure out the "mystery" here?

tadzik: I'd love to, but I can't make sense of it: it keeps being text (as it should be), but somehow, the moment you click on it, it becomes the recipient object for a single tick, and for that single tick the searchformcontrol puts it through here.

petersalomonsen: could it be because you converted the search index suggestions to recipient objects? In that case I guess it's ok.

tadzik: Yeah, that's probably that. Weird how it works alright in all other cases except right when you click them though :)

petersalomonsen: The search text field will be a string, but the auto-complete popup results are Recipient objects. So maybe the appropriate would be to change line 66 to something like:

`.subscribe((searchtext: string | Recipient) => {`

tadzik: Ah, that's sensible. Will do :)

PR #156, opened 2019-05-16
Tadzik/calendar/import export

tadzik:


tadzik:
(Rebased on top of new master now)

PR #157, opened 2019-05-19
Canvastable drag select

petersalomonsen:
Fixing an annoyance when selecting multiple rows, that the first clicked row was not selected. Also handling that rows in between were not selected if moving mouse fast. Now also possible to undo selection by dragging back up again.

Added end to end test too.

multiselectrunbox7

### DIFF
--- a/src/app/canvastable/canvastable.ts
+++ b/src/app/canvastable/canvastable.ts
@@ -234,6 +234,8 @@ export class CanvasTableComponent implements AfterViewInit, DoCheck {
     }
   }
 
+  private dragSelectionDirectionIsDown: Boolean = null;
+
   // Auto row wrap mode (width based on iphone 5) - set to 0 to disable row wrap mode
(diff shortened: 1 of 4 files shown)


src/app/canvastable/canvastable.ts
```diff
@@ -233,6 +233,8 @@ export class CanvasTableComponent implements AfterViewInit, DoCheck {
     }
   }
 
+  private dragSelectDirection: Boolean = null;
```
tadzik: I assume this is up/down; maybe it could be either commented or named to indicate what is what? Perhaps `dragSelectionIsDown` or something along those lines?

petersalomonsen: Good point! Changed to `dragSelectionDirectionIsDown`.

tadzik: Excellent :)

PR #158, opened 2019-05-20
Tadzik/contacts/contacts being orgs

tadzik:


tadzik:
This requires 48ee59ccf7 to be deployed on the backend first, so on hold for now.

PR #164, opened 2019-05-21
Tadzik/calendar/recurring events

tadzik:


src/app/calendar-app/event-editor-dialog.component.ts
```diff
@@ -62,6 +77,18 @@ export class EventEditorDialogComponent {
...
     onSubmitClick(): void {
+        this.event.dtstart = moment(this.event.start);
+        this.event.dtend = moment(this.event.end);
+
+        if (this.recurring_frequency === -1) {
+            if (this.event.rrule) {
+                this.event.rrule = undefined;
+            }
+        } else {
+            // FIXME preserve possible existing other settings!
```
petersalomonsen: Is this a reminder for later?

tadzik: Oh, so that's why I didn't submit that PR before... ;)

Thanks for catching it, indeed it is.

tadzik:
> A commit about "Making linter happy" is not so interesting on the master branch. Also this could be one commit with `calendar(feat): recurring events` in the subject line, and the individual commit comments as lines in the commit message body parts.

I agree about the linter, it could indeed be squashed into something.

However, I'm firmly against merging several commits into one for the sake of preserving the niceness of the master commit log. The reason why commits are made separately and as granular as possible is to make it easier to bisect, or in other ways locate something that introduced a bug in the future. Commits like https://github.com/runbox/runbox7/pull/164/commits/cd3bdf173b776673b99ede9759b6d454820d28f2 or https://github.com/runbox/runbox7/pull/164/commits/0351e1dddddc6040a8770b291dfc323658a1552f are not strictly related to recurring events, they could've as well been commited separately as bugfixes. A mega-commit with hundreds of lines in it concerning different things is not useful for future reference, and I'd argue it's not very useful as a changelog entry either.

I understand the desire to keep the commit log clean as much as possible, but perhaps it could be achieved with a clear naming scheme for commits: [FEATURE], [CHANGE], [BUGFIX] or something along those lines without losing the good things that version control gives us.

gtandersen:
> I understand the desire to keep the commit log clean as much as possible, but perhaps it could be achieved with a clear naming scheme for commits: [FEATURE], [CHANGE], [BUGFIX] or something along those lines without losing the good things that version control gives us.

We have established some commit types in https://github.com/runbox/runbox7/blob/master/CONTRIBUTING.md#type. :)

tadzik:
> > I understand the desire to keep the commit log clean as much as possible, but perhaps it could be achieved with a clear naming scheme for commits: [FEATURE], [CHANGE], [BUGFIX] or something along those lines without losing the good things that version control gives us.
> 
> We have established some commit types in https://github.com/runbox/runbox7/blob/master/CONTRIBUTING.md#type. :)

Yeah, I learned about those recently, later than I should have ;) I've rebased the branch, squashed the irrelevant commits and reworded the existing ones so that they fit the coding standard properly.

src/app/calendar-app/event-editor-dialog.component.ts
tadzik: Fixed now in cf93f06

PR #188, opened 2019-06-04
fix: Improve layout and formatting of Contacts screens

gtandersen:
Adjustments to left pane, textual content, and buttons.

Screen Shot 2019-06-04 at 23 23 57

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -104,7 +104,27 @@ h5 {
     display: inline !important;
 }
 
+mat-icon {
+    margin-bottom: 3px;
+    vertical-align: middle !important;
+}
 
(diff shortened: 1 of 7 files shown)


src/styles.scss
```diff
@@ -104,7 +104,28 @@ h5 {
...
+    line-height: 1.5em !important;
+    text-decoration: none !important;
+    white-space: nowrap;
+}
+
+.mainTextContent {
+    padding: 3em;
+}
 
+.contentButton mat-icon {
```
tadzik: What does an empty block like this do? Does it "reset" the things set in the regular `.contentButton`?

PR #191, opened 2019-06-11
fix: Improve layout and formatting of Month view and Add Event view.

gtandersen:
- Even calendar columns.
- Decrease event font size to fit more text in each square.
- Decrease sidenav width.
- Increase title font size.
- Decrease Add Event height.

Screen Shot 2019-06-11 at 12 14 24

### DIFF
--- a/src/app/calendar-app/calendar-app.component.html
+++ b/src/app/calendar-app/calendar-app.component.html
@@ -94,7 +94,7 @@
         </mat-toolbar>
 
         <ng-template #dayViewTemplate let-day="day" let-locale="locale">
-            <div>
+            <div class="calendarMonthWeek">
                 <div class="cal-cell-top">
                     <span class="cal-day-number">
                         {{ day.date | calendarDate:'monthViewDayNumber':locale }}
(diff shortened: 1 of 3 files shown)


tadzik:
Looks good from the screenshots but I tried it locally and found a few things to pick on :­P

The un-hardcoding of widths and heights (https://github.com/runbox/runbox7/pull/191/files#diff-7ffe951f0eb9eb4cab2d6fc7781e5d36L97) makes things more dynamic which can be a good thing, but it makes days change sizes depending on the number of events in them:

Before:
2019-06-11-094414_274x340_scrot

After:
2019-06-11-094520_271x275_scrot

It looks good on your screen, I think the difference may come from the use of pixels in styles: on different screens things may scale up differently because of that. I tried using em/pt/vmin before for consistency, perhaps it'd be good to switch to them here too?

gtandersen:
Thanks @tadzik -- how does this look now:

Screen Shot 2019-06-12 at 14 06 18

tadzik:
Cell formatting looks good now :) But I'm still a bit bothered by the font size: pixels have different sizes on different screens, and on my FHD 14" laptop screen the font size is super tiny:
2019-06-12-110923_1916x1055_scrot

Could you make it use `pt` so that it scales according to the DPI?

gtandersen:
The pt unit is traditionally used in print media and not for web/CSS because it's unreliable across browsers and screens. All text in your screenshot appears smaller than average due to the high resolution, which means that more events will fit each cell. With an average resolution, 3-4 events will fit and I think that's a good compromise.

PR #198, opened 2019-06-13
refactor(Calendar): split a calendar service from a calendar component

tadzik:
This will enable calendar interactions from the rest of the app, including adding events from email attachments.


tadzik:
Pushed some changes to adapt tests to the refactor

PR #203, opened 2019-06-18
docs: lint and test instructions

petersalomonsen:
add instructions for lint and tests
update TOC

@tadzik We probably need instructions for running the license check scripts too?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,14 +4,24 @@ We would love for you to contribute to Runbox 7 and help make it even better tha
 
 As a contributor, here are the guidelines we would like you to follow:
 
-* [Code of Conduct](#code-of-conduct)
-* [Question or Problem?](#got-a-question-or-problem)
-* [Issues and Bugs](#found-a-bug)
-* [Feature Requests](#missing-a-feature)
-* [Submission Guidelines](#submission-guidelines)
-* [Coding Rules](#coding-rules)
-* [Commit Message Guidelines](#commit-message-guidelines)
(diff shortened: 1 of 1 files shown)


tadzik:
Perhaps it's best if we just integrate the license tests with the rest of the tests. I also noticed that right now they contain a non-core Perl dependency (`Path::Tiny`) so it'd even need separate things installed to be runnable.

Maybe it's better if we rewrite it in JS and make it run with the rest of the test suite.

petersalomonsen:
yes maybe better to have it in JS if that's an easy fix @tadzik. Also considering recommending to set up travis-ci for the fork/branch before it's submitted as a PR. This can compensate for not running tests locally. What do you think? Have you set up a separate Travis CI for your fork? (shouldn't be more than registering your github account with https://travis-ci.com as far as I remember)

tadzik:
Notes from yesterday's meeting:

– let's rewrite license tests in JS so that Perl is no longer a dependency and the tests themselves run along with `npm run test`, so that they don't need further explaining

– let's not compensate for running tests locally but rather make it easier to run tests locally – so that something like `npm run verify` runs `test`, `list`, `e2e` and `build` just like travis

These two should probably be done separately to not delay these fixes here.

tadzik:
We can revisit this now that `npm run policy` is a thing.

tadzik:
Forcepushed to reword your commit, now that policy tests require it :)

tadzik:
I integrated the TOC changes in #373, now that the (updated) test instructions have been added.

Issue #45, opened 2019-01-10
Add sender to Contacts


gtandersen:
Functionality for adding the sender of an email to Contacts, similar to Runbox 6.

Ref: https://community.runbox.com/t/adding-contacts-from-to-address/716

PR #205, opened 2019-06-19
feat(mailviewer): Make email addresses clickable

tadzik:
Froms, Tos, CCs etc are now clickable, enabling the user to show the
contacts they're related to (if they exist) or add new ones.

Fixes GH-45.

Looks like this – dotted line for new contacts, solid for existing ones.
2019-06-19-132743_517x125_scrot

### DIFF
--- a/src/app/contacts-app/contact-details/contact-details.component.ts
+++ b/src/app/contacts-app/contact-details/contact-details.component.ts
@@ -64,6 +64,21 @@ export class ContactDetailsComponent {
             const contactid = params.id;
             if (contactid === 'new') {
                 this.contact = new Contact({});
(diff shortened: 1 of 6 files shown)


tadzik:
I'll give it a try; I initially went with just the icons, but it ended up being quite noisy on the screen with emails that had lots of CCs in them. Those icons look a lot better than the ones I tried though­ – I'll give them a shot :)

tadzik:
I think it's looking pretty nice :)

2019-06-20-124243_528x76_scrot

Perhaps a bit noisy when there's a lot of recipients, but we can always reduce it a little if it becomes problematic – and now it definitely stands out more and will bring some attention to the feature :)

PR #218, opened 2019-07-05
test(policy): add tests for commit messages

tadzik:
Along with license comment tests they shall now be called "policy tests"
and ran together under that name.


tadzik:
Seems to run fine: https://travis-ci.com/runbox/runbox7/builds/118095844#L588 :)

Issue #190, opened 2019-06-08
Folder message counts include messages marked for deletion


davidbowdley:
The folder message counts include messages marked for deletion but not purged yet. 

Notably this is a problem with Apple Mail, but also clients also have this feature such as Thunderbird. Messages that are deleted or moved are marked for deletion but not purged immediately. 

In RMM6 we show these with a strike-through as technically they are still present in the mailbox and taking up storage. However, this leads to confusion with users because other webmail services don't show these messages. 

In Runbox 7 we decided to not show these messages, however the folder count still includes them which means the folder count appears to be incorrect when compared to the messages you can actually see in the folder.

PR #231, opened 2019-08-13
Sync index on foldercount mismatch

petersalomonsen:
This PR adds comparing folder counts received from the server (which is displayed in the left side menu), with the actual count per folder in the local search index. If there are discrepancies, the current folder in view will be retrieved from the server, and added to the index. If the next compare still show discrepancies, a dialog will pop up and suggest downloading a fresh index from the server.

A cause of discrepancies is that the folder count from the server includes messages marked for deletion ( as mentioned in https://github.com/runbox/runbox7/issues/190 ). To prevent this we should decide on either supporting strikethrough of messages marked from deletion in Runbox7, or remove them from the counters as suggested in the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,11 @@ cache:
      - ./node_modules
 
 install:
-  - npm install
-  - npm run webdriver-update-ci
+  - npm install  
 
 script:
   - npm run lint
   - npm run policy
   - npm run test -- --watch=false --progress=false --browsers=ChromeHeadlessCI
-  - npm run e2e -- --webdriver-update=false --protractor-config=./protractor-ci.conf.js
+  - npm run e2e -- --protractor-config=./protractor-ci.conf.js
   - npm run build
--- a/package-lock.json
+++ b/package-lock.json
@@ -1121,9 +1121,9 @@
       "dev": true
     },
     "@types/selenium-webdriver": {
-      "version": "3.0.14",
-      "resolved": "https://registry.npmjs.org/@types/selenium-webdriver/-/selenium-webdriver-3.0.14.tgz",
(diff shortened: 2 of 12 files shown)


gtandersen:
@petersalomonsen The first change which compares folder counts only have benefits AFAICT.

Regarding strike-throughs then supporting this in user interfaces seems to be uncommon, so removing them from the folder counts appears to be the best solution.

petersalomonsen:
ok, for the folder count to exclude messages marked for deletion, that would have to be on the server.

Also another issue I found after testing this was that while it does download all message headers for a folder on discrepancies, and add them to the indexer again, it does not remove those that are deleted or moved to another folder - so you could end up with too many messages in the local index for that folder. I'll try to come up with something there ( so still not ready for merge ).

also have to fix the commit messages.

petersalomonsen:
The latest commit ( https://github.com/runbox/runbox7/pull/231/commits/4ff6c19 ) adds a check for messages that are in the local index but not on the server, and sends a verification requests for these messages which will either update the changed_date or deleted timestamp, which then again will be picked up on next check for messages to be synced from the server.

What still remains is to handle messages marked for deletion (deleted flag), and since we're not going to show these (with strikethrough), a solution might be that the Runbox7 app actually deletes these messages from the server. Will that be a problem ? ( @davidbowdley you might have an opinion on this ).

petersalomonsen:
Latest commit shows messages marked for deletion with a unicode recycle symbol (didn't find trash), also a tooltip explaining why.

Runbox7markdelete

petersalomonsen:
Also added a missing DKIMModule import ( https://github.com/runbox/runbox7/pull/231/files#diff-fc6b22833aeaf77f8440c4970ade705eR136 )  that caused the tests to fail from this commit https://github.com/runbox/runbox7/commit/babd3dcf1549276cc9eb1612e10be320b49aaefb @hernan604

gtandersen:
Looks good @petersalomonsen !

PR #248, opened 2019-08-20
refactor(email): Split the webmail into a separate app

tadzik:
Because it's now no longer the "main" app, and only loaded when it's
actually needed it won't interfere with the usage of Runbox7 unrelated
to emails themselves: for example allowing to renew the account through
payments-app without the searchservice and checks for new messages
interfering and failing due to an expired account.


petersalomonsen:
This looks very promising @tadzik !

PR #253, opened 2019-08-21
dkim reconfigure cnames option

hernan604:


src/app/dkim/dkim.component.ts
```diff
@@ -134,6 +134,27 @@ export class DkimComponent implements AfterViewInit {
     }
   }
 
+  reconfigure (item) {
+    const url = '/rest/v1/dkim/' + this.domain + '/reconfigure/' + item.selector;
+    const req = this.http.put(url, {});
```
tadzik: Any reason why this is not in rmmapi?

PR #254, opened 2019-08-26
feat(folderlist): drag and drop reordering

### LINKED ISSUE #34: Functionality to rearrange folders
Functionality to drag & drop folders and thereby rearrange the folder hierarchy.

petersalomonsen:
**NOTE:** While the UI updates here supports reordering of the folders, the backend does not. Changing the order of folders will be immediately reflected in the user interface, but will be reset when a new folderlist is read from the server. We need backend support for setting the order of folders.

- add drag and drop reordering of folders
- fix canvas table broken drag messages to folder (was broken in 7dc8150e)
- indicate drop above / below / inside when dragging over
- move above / below / into folder ( **IMPORTANT: Backend currently only supports moving into folders, not above/below)**
- add unit test for various cases of rearranging folders
- add unit test for canvastable draggable overlay being activated on hover
- changed to using dragend event instead of dragleave
- load xapian only on demand

fixes #34

### DIFF
--- a/src/app/canvastable/canvastable.ts
+++ b/src/app/canvastable/canvastable.ts
@@ -494,13 +494,16 @@ export class CanvasTableComponent implements AfterViewInit, DoCheck {
         const colIndex = this.getColIndexByClientX(clientX);
         let colStartX = this.columns.reduce((prev, curr, ndx) => ndx < colIndex ? prev + curr.width : prev, 0);
 
-        let tooltipText = this.columns[colIndex] && this.columns[colIndex].tooltipText;
+        let tooltipText: string | ((rowobj: any) => string) =
+              this.columns[colIndex] && this.columns[colIndex].tooltipText;
 
-        if (typeof tooltipText === 'function') {
+        if (typeof tooltipText === 'function' && this.rows[this.hoverRowIndex]) {
           tooltipText = tooltipText(this.rows[this.hoverRowIndex]);
         }
 
-        if (!event.shiftKey && !this.lastMouseDownEvent && tooltipText) {
+        if (!event.shiftKey && !this.lastMouseDownEvent &&
+            (tooltipText || (this.columns[colIndex] && this.columns[colIndex].draggable))
+          ) {
           if (this.rowWrapMode &&
             colIndex >= this.rowWrapModeWrapColumn) {
             // Subtract first row width if in row wrap mode
@@ -512,7 +515,7 @@ export class CanvasTableComponent implements AfterViewInit, DoCheck {
             (this.hoverRowIndex - this.topindex) * this.rowheight,
             colStartX - this.horizScroll + this.colpaddingleft,
             this.columns[colIndex].width - this.colpaddingright - this.colpaddingleft,
-            this.rowheight, tooltipText);
+            this.rowheight, tooltipText as string);
 
           if (this.rowWrapMode) {
             this.floatingTooltip.top +=
--- a/src/app/folder/folderlist.component.html
+++ b/src/app/folder/folderlist.component.html
@@ -5,15 +5,19 @@
   </button>
 </div>
 <mat-tree [dataSource]="dataSource" [treeControl]="treeControl">
-    <mat-tree-node *matTreeNodeDef="let node" matTreeNodePadding
-      matTreeNodePaddingIndent="20" style="display: flex"
+    <mat-tree-node *matTreeNodeDef="let node" [matTreeNodePadding]="node.folderLevel"
+      matTreeNodePaddingIndent="20"
+      style="display: flex"
+      class="mailFolder"
       [ngClass]="{
         'selectedFolder' : selectedFolder == node.folderPath,
-        'dropToFolderAllowed' : dropFolderId === node.folderId
+        'dropToFolderAllowed' : dropFolderId===node.folderId && dropAboveOrBelowOrInside === dropPosition.INSIDE,
+        'dropAboveFolderAllowed' : dropFolderId===node.folderId && dropAboveOrBelowOrInside === dropPosition.ABOVE,
+        'dropBelowFolderAllowed' : dropFolderId===node.folderId && dropAboveOrBelowOrInside === dropPosition.BELOW
         }"
-        (drop)="dropToFolder(node.folderId)"         
-        (dragover)="allowDropToFolder($event,node.folderId);treeControl.expand(node)" 
-        (dragleave)="dropFolderId=null" class="mailFolder"
+        (drop)="dropToFolder($event,node.folderId)"         
+        (dragover)="allowDropToFolder($event,node)" 
+        (dragend)="dragCancel()" 
         (click)="selectFolder(node.folderPath)"
       >
         <button mat-icon-button matTreeNodeToggle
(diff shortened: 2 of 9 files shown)


src/app/canvastable/canvastable.ts
```diff
@@ -512,7 +515,7 @@ export class CanvasTableComponent implements AfterViewInit, DoCheck {
             (this.hoverRowIndex - this.topindex) * this.rowheight,
             colStartX - this.horizScroll + this.colpaddingleft,
             this.columns[colIndex].width - this.colpaddingright - this.colpaddingleft,
-            this.rowheight, tooltipText);
+            this.rowheight, tooltipText as string);
```
tadzik: I think it's guaranteed to be a string by now (thanks to L500), so this is just to make Typescript happy?

src/app/folder/folderlist.component.html
```diff
@@ -30,9 +34,13 @@
...
         <mat-icon *ngIf="node.folderType=='user'" mat-list-icon class="folderIconUser">folder</mat-icon>
-	<span style="margin-left: 5px;">{{node.folderName}}</span>
+        <div style="margin-left: 5px;" 
+              draggable="true"
+              (dragstart)="dragFolderStart($event, node.folderId)">
+              {{node.folderName}}
+        </div>
         <span *ngIf="node.newMessages" matBadge="{{node.newMessages}}" matBadgeOverlap="true" class="newMessagesCount" matBadgeSize="medium">&nbsp;</span>
-	<span style="flex-grow: 1"></span>
+	      <span style="flex-grow: 1"></span>
```
tadzik: Extra whitespace? :­)

src/app/folder/folderlist.component.ts
```diff
@@ -43,6 +44,8 @@ class FolderNode {
 export class FolderListComponent {
     selectedFolder = 'Inbox';
     dropFolderId: number;
+    dropAboveOrBelowOrInside = 0;
```
tadzik: Wouldn't this be better off as an [enum](https://www.typescriptlang.org/docs/handbook/enums.html)?

src/app/folder/folderlist.component.ts
```diff
@@ -140,13 +144,58 @@ export class FolderListComponent {
     private _getLevel = (node: FolderCountEntry) => node.folderLevel;
     private _isExpandable = (node: FolderCountEntry) => node.isExpandable ? true : false;
 
-    allowDropToFolder(event, folderId: number): void {
+    /**
+     * Folderlist entry is 48 pixels, so if mouse is in the upper region suggest dropping above,
```
tadzik: 48 is a bit of a magic number here.

Is it possible to fetch this number of pixels straight from the Folderlist component? This would make us more future-proof.

src/app/rmmapi/rbwebmail.ts
```diff
@@ -337,6 +337,15 @@ export class RunboxWebmailAPI {
         return req.pipe(map((res: any) => res.status === 'success'));
     }
 
+    moveFolder(folderId: number, newParentFolderId: number): Observable<boolean> {
+        const req = this.http.put('/rest/v1/email_folder/move', {
+            'to_folder': newParentFolderId,
+            'folder_id': folderId
+        });
+        this.subscribeShowBackendErrors(req);
```
tadzik: I had no idea we have this! Why aren't we using it everywhere? :­D

src/app/canvastable/canvastable.ts
petersalomonsen: Yes typescript was not convinced that this had to be a string at this point ( it seems normally quite intelligent on these matters, but not in this case ).

src/app/folder/folderlist.component.html
petersalomonsen: yes, previous revision only had 1 tab on this line, while the ident level in that block is 4 tabs.

src/app/folder/folderlist.component.ts
petersalomonsen: changed to using `enum` now.

src/app/folder/folderlist.component.ts
petersalomonsen: Now getting the tree node height from DOM.

src/app/folder/folderlist.component.ts
```diff
@@ -43,6 +50,8 @@ class FolderNode {
 export class FolderListComponent {
     selectedFolder = 'Inbox';
     dropFolderId: number;
+    dropAboveOrBelowOrInside: DropPosition = 0;
```
tadzik: Shouldn't this be `... = DropPosition.NONE`?

src/app/folder/folderlist.component.ts
tadzik: This makes the comment outdated – but it's great! Is it possible to do the same for the `7`?

src/app/folder/folderlist.component.ts
```diff
@@ -220,6 +277,120 @@ export class FolderListComponent {
...
+            // can't move a folder above, below or inside itself
+            return;
+        }
+
+        const folders = await this.messagelistservice.folderCountSubject.pipe(take(1)).toPromise();
+        let sourceIndex = folders.findIndex(fld => fld.folderId === sourceFolderId);
+        let destinationIndex = folders.findIndex(folder => folder.folderId === destinationFolderId);
+
+        console.log(`move folder ${folders[sourceIndex].folderPath} ${
+                            aboveOrBelowOrInside === 1 ? 'above' :
```
tadzik: We don't seem to be using the enum names here, is there a reason?

src/app/folder/folderlist.component.ts
```diff
@@ -220,6 +277,120 @@ export class FolderListComponent {
...
+
+        let moveCount = 1;
+        while ( sourceIndex + moveCount < folders.length &&
+                folders[sourceIndex + moveCount].folderLevel > folders[sourceIndex].folderLevel) {
+            // also move all sub folders if any
+            moveCount ++;
+        }
+
+        switch (aboveOrBelowOrInside) {
+            case 1:
```
tadzik: Not using enums either.

src/app/folder/folderlist.component.html
```diff
@@ -5,15 +5,19 @@
...
-    <mat-tree-node *matTreeNodeDef="let node" matTreeNodePadding
-      matTreeNodePaddingIndent="20" style="display: flex"
+    <mat-tree-node *matTreeNodeDef="let node" [matTreeNodePadding]="node.folderLevel"
+      matTreeNodePaddingIndent="20"
+      style="display: flex"
+      class="mailFolder"
       [ngClass]="{
         'selectedFolder' : selectedFolder == node.folderPath,
-        'dropToFolderAllowed' : dropFolderId === node.folderId
+        'dropToFolderAllowed' : dropFolderId===node.folderId && dropAboveOrBelowOrInside === 3,
```
tadzik: The checks don't seem to be using enum names – is there a reason to not use them in HTML?

src/app/folder/folderlist.component.ts
petersalomonsen: No, since that's a more like a choice we have to make. Dropping within the 7 upper or lower will result in the folder being dropped above or below. So that's nothing that can be read from an element. It's more like a constant to be adjusted if the space for drop above/below needs to be increased / reduced.

src/app/folder/folderlist.component.ts
petersalomonsen: Indeed it should :+1:  Fixed now!

src/app/folder/folderlist.component.ts
petersalomonsen: Thanks! I missed that one.

src/app/folder/folderlist.component.ts
petersalomonsen: ohh... I hope I've corrected all of those now :-) Thanks for finding these!

src/app/folder/folderlist.component.html
petersalomonsen: No - it's more readable using them in HTML too, so I just had to find out how. I used the method described here: https://stackoverflow.com/questions/35835984/how-to-use-a-typescript-enum-value-in-an-angular2-ngswitch-statement/35835985#35835985

Looks better now?

src/app/folder/folderlist.component.ts
tadzik: Alright, makes sense :)

src/app/folder/folderlist.component.html
tadzik: Lovely!

PR #268, opened 2019-08-28
feat(account-app): Add an account-app implementing payments interface

tadzik:


src/app/account-app/account-addons.component.html
```diff
@@ -0,0 +1,58 @@
...
+
+.productCard {
+	margin: 10px;
+	flex: 600px 1 0;
+}
+</style>
+
+<h1> Sub-accounts </h1>
+
+<p> Some blurb about why this is cool </p>
```
petersalomonsen: someone will author the "blurb"?

src/app/account-app/account-app.module.ts
```diff
@@ -0,0 +1,172 @@
...
+  entryComponents: [
+    BitpayPaymentDialogComponent,
+    PaypalPaymentDialogComponent,
+    StripePaymentDialogComponent,
+  ],
+  providers: [
+    PaymentsService,
+    ScriptLoaderService,
+  ],
+  bootstrap: [AccountAppComponent]
```
petersalomonsen: Why bootstrapping ? are you running this standalone in some cases?

src/app/account-app/account-components.component.ts
```diff
@@ -0,0 +1,35 @@
...
+import { Component } from '@angular/core';
+
+import { RunboxMe, RunboxWebmailAPI } from '../rmmapi/rbwebmail';
+
+@Component({
+    selector: 'app-account-components-component',
+    templateUrl: './account-components.component.html',
+})
+export class AccountComponentsComponent {
+    me: RunboxMe = new RunboxMe();
```
petersalomonsen: cant't see any reference to "me" in `account-components.component.html`. Are you using this?

src/app/account-app/account-product.component.ts
```diff
@@ -0,0 +1,86 @@
...
+
+import { Component, Input } from '@angular/core';
+import { PaymentsService } from './payments.service';
+import { Product } from './product';
+import { ProductOrder } from './product-order';
+
+@Component({
+    selector: 'app-account-product',
+    template: `
+<mat-card>
```
petersalomonsen: Would it be better to separate this in its own html file?

src/app/account-app/account-receipt.component.ts
```diff
@@ -0,0 +1,53 @@
...
+import { ActivatedRoute } from '@angular/router';
+import { RunboxMe, RunboxWebmailAPI } from '../rmmapi/rbwebmail';
+
+@Component({
+    selector: 'app-account-receipt-component',
+    templateUrl: './account-receipt.component.html',
+})
+export class AccountReceiptComponent {
+    receipt: any = { products: [] };
+    me: RunboxMe = new RunboxMe();
```
petersalomonsen: in this case you will overwrite me as soon as the subcribe is done. I'm wondering if it would be better to use an asyncsubject here instead of starting with an empty RunboxMe object. In fact `me` is already an async subject that you can use directly from here: https://github.com/runbox/runbox7/blob/master/src/app/rmmapi/rbwebmail.ts#L166

src/app/account-app/account-transactions.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    methods = {
+        bitpay:     'Bitpay',
+        creditcard: 'Netaxept',
+        giro:       'Offline',
+        paypal:     'PayPal',
+        stripe:     'Stripe',
+    };
+
+    statuses = {
+        0: 'Successful',
```
petersalomonsen: Would "string enums" make any sense here? https://www.typescriptlang.org/docs/handbook/enums.html

src/app/account-app/account-upgrades.component.ts
```diff
@@ -0,0 +1,45 @@
...
+import { RunboxMe, RunboxWebmailAPI } from '../rmmapi/rbwebmail';
+import { PaymentsService } from './payments.service';
+import { Product } from './product';
+
+@Component({
+    selector: 'app-account-upgrades-component',
+    templateUrl: './account-upgrades.component.html',
+})
+export class AccountUpgradesComponent {
+    me: RunboxMe = new RunboxMe();
```
petersalomonsen: Same as notice for `me` above

src/app/account-app/cart.ts
```diff
@@ -0,0 +1,59 @@
...
+// along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
+// ---------- END RUNBOX LICENSE ----------
+
+import { ProductOrder } from './product-order';
+
+export class Cart {
+    items: ProductOrder[] = [];
+
+    constructor() {
+        const storedCart = localStorage.getItem('shoppingCart');
```
petersalomonsen: would it be an idea to prepend / append user account id? if using multiple accounts in the same browser.

src/app/account-app/payments.service.ts
```diff
@@ -0,0 +1,65 @@
...
+    stripePubkey   = new ReplaySubject<string>(1);
+    currency       = new ReplaySubject<string>(1);
+
+    cart: Cart = new Cart();
+
+    constructor(
+        private rmmapi: RunboxWebmailAPI,
+    ) {
+        this.rmmapi.getAvailableProducts().subscribe(products => {
+            this.products.next(products);
```
petersalomonsen: ReplaySubject or AsyncSubject? If there's only one event to be pushed here in the app lifetime, I would probably rather use a completing AsyncSubject.

src/app/account-app/paypal-payment-dialog.component.ts
```diff
@@ -0,0 +1,72 @@
...
+        private router: Router,
+        public dialogRef: MatDialogRef<PaypalPaymentDialogComponent>,
+        @Inject(MAT_DIALOG_DATA) public data: any
+    ) {
+        this.tid = data.tx.tid;
+        // not using Router here because we want the entire URL, hostname and all
+        const return_url = window.location.href.replace(/account.*/, 'account/paypal/confirm');
+        const cancel_url = window.location.href.replace(/account.*/, 'account/paypal/cancel');
+        this.rmmapi.payWithPaypal(this.tid, return_url, cancel_url).subscribe(res => {
+            console.log(res);
```
petersalomonsen: debugging leftovers?

src/app/account-app/scriptloader.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+// adopted from https://stackoverflow.com/a/42766146
+@Injectable()
+export class ScriptLoaderService {
+    scripts = {
+        stripe: { loaded: false, src: 'https://js.stripe.com/v3/' }
+    };
+
+    loadScript(name: string): Promise<void> {
+        return new Promise((resolve, reject) => {
```
petersalomonsen: I think probably using an asyncsubject for script loading is better than promise and maintaining a tracking object. Like the scriptloader I've made for loading Xapian: https://github.com/runbox/runbox7/blob/master/src/app/xapian/xapianwebloader.ts#L37

src/app/account-app/account-addons.component.html
tadzik: I think @gtandersen had some ideas :)

src/app/account-app/account-app.module.ts
tadzik: Hm, I don't think so. Will remove it.

tadzik: Fixed in 0d0e053, will squash them later.

src/app/account-app/account-components.component.ts
tadzik: I don't think so, looks like a leftover. Fixed in 3c3b6f1.

src/app/account-app/account-product.component.ts
tadzik: I was about to say “but it's only a few lines of code”, but then I saw that it's not anymore. They grow up so fast! Fixed in a0d3b9e :)

src/app/account-app/account-receipt.component.ts
tadzik: I think I put it here as an empty object so that the HTML rendering doesn't break on an undefined variable when I do `{{ me.username }}` and alike in the template.

“Make sure the template isn't broken before the data arrives” is a common problem I have – how would I solve it elegantly here?

src/app/account-app/cart.ts
tadzik: Seems to be a recurring theme too: I had the same thing (and you had the same comment :)) in the offline calendar PR. Makes me think that perhaps we should have some common Service that handles the offline stuff, so it's harder to forget to do that?

src/app/account-app/payments.service.ts
tadzik: Good question; I think 3/4 of those are “immutable”, while `activeProducts` may change (whenever you buy something). That said, we never refresh it anywhere, so there is currently no point in them being `ReplaySubject`s. So if I use `AsyncSubject` here I should also do the `.complete()` later – but will future subscribers still get the value that we obtained here?

src/app/account-app/paypal-payment-dialog.component.ts
tadzik: Indeed. Fixed in fb181d1.

src/app/account-app/scriptloader.service.ts
tadzik: `AsyncSubject` may indeed make more sense. I created the tracking object because I thought that we might need this kind of dynamic-external-JS-loading in other places too. Now I see that we also needed it for xapian, and that my `scriptloader` is not quite good enough for it needs either :P So probably a YAGNI on my part, will remove it.

tadzik: Addressed in e88c0f8.

src/app/account-app/account-receipt.component.ts
tadzik: Done in 17c87e3.

src/app/account-app/account-upgrades.component.ts
tadzik: Fixed now.

src/app/account-app/payments.service.ts
tadzik: Addressed in fd65e0b.

src/app/account-app/cart.ts
tadzik: It required a bit of a redo of the cart and everything that uses it (ie the entire account app), but it's up now: https://github.com/runbox/runbox7/commit/0a71f7783c7fc3f1aea729574fb4d8b49c24b736.

It touches a lot of things and hopefully doesn't break too many of them. There will be tests for it too, but I'll appreciate a good luck at the async stuff, it was a bit of a PITA to get right.

src/app/account-app/account-addons.component.html
tadzik: Fixed now.

PR #273, opened 2019-08-30
fix(payments): Improve layout, formatting, text, and navigation.

gtandersen:


tadzik:
The commits are now in, but reworded and github didn't pick it up. Thanks!

PR #281, opened 2019-09-01
fix(dev): Don't check security with custom development backend.

gtandersen:


backend-proxy-remote.conf.js
```diff
@@ -1,5 +1,10 @@
 let backendhostname = process.env.RUNBOX7_ANGULAR_BACKEND_HOST;
-if(!backendhostname) {
+if(backendhostname) {
+    // If we're using a custom development backend, turn off security checking to avoid certificate errors:
+    backendhostsecure = false;
```
tadzik: I think now we never have it set to `true`: it's either `false` from this is or undefined – which is probably also false.

Maybe it should be predeclared up top, after backendhostname? It can be false initially, then set to true if we get to the default production case.

backend-proxy-remote.conf.js
```diff
@@ -27,7 +32,7 @@ const PROXY_CONFIG = [
             }
         },
         "target": backendhostname,
-        "secure": true,
+        "secure": backendhostsecure,
```
tadzik: We should do the same in the websocket section.

gtandersen: Thanks @tadzik; should be fixed now.

tadzik:
Looks like it fails the commit message guidelines: https://travis-ci.com/runbox/runbox7/builds/125978096#L707 :(

tadzik:
Looks good, but the commit messages need fixing.

PR #283, opened 2019-09-02
test(calendar): handle beginning of month

petersalomonsen:
add 15 days to the second calendar test event so that it's not visible
in month view in the beginning of a month


petersalomonsen:
saw @tadzik already made a PR for this. closing..

PR #286, opened 2019-09-05
fix(folder): make sure folders are handled correctly regadless of order

tadzik:
With the new priority sorting (done by getFolderCount() in RunboxWebmailAPI)
there is a risk that subfolders appear before parent folders in the
resulting list. The flattening algorithm in FolderListComponent doesn't
handle that correctly, breaks and displays nothing. This fixes it and
adds a testcase.

It feels like now that the folders are pre sorted the BFS loop (or what
looks like one :)) below could be somewhat simplified too – but that's
a task for another time.

### DIFF
--- a/src/app/folder/folderlist.component.spec.ts
+++ b/src/app/folder/folderlist.component.spec.ts
@@ -282,4 +282,16 @@ describe('FolderListComponent', () => {
         expect(rearrangedFolders.map(f => f.folderLevel)).toEqual([0, 0, 0, 0, 1, 2, 1]);
         expect(ordered_ids_request).toEqual([1, 7, 6, 5, 3, 2, 4]);
     });
+    it('folders provided out of order should work fine', async () => {
+        const comp = new FolderListComponent({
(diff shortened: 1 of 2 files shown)


petersalomonsen:
would it be more appropriate to prevent folder reordering to put subfolders above their parents? 
You could also add a testcase for such a scenario here here: https://github.com/runbox/runbox7/blob/330b926d27593f4d9ad6d863e54a824d053e0ada/src/app/folder/folderlist.component.spec.ts#L276

tadzik:
Ah, I see what you mean now.

Well, I did manage to break it *somehow*, but I have no idea how. I was playing around with it and suddenly it's broken, so I can't even think of a testcase for it. Seems to be a bit elusive too, given the fact that I seem to be the only one with that problem so far :/

tadzik:
Given that I have no idea how to reproduce the real bug, and this is actually stopping me from using folders (and maybe it's not just me :)) on production for the last few weeks, would you be okay with merging this?

PR #305, opened 2019-09-23
Try to split travis build into smaller parts

tadzik:


petersalomonsen:
This is great @tadzik Much better to have separate travis jobs.

PR #321, opened 2019-10-13
test(e2e): add tests for folder creation at root level and email title changes to subject text

WillEllis:
Add an e2e test for checking the email title changes to the subject text & test for creating folder at root level

### DIFF
--- a/e2e/compose/compose.e2e-spec.ts
+++ b/e2e/compose/compose.e2e-spec.ts
@@ -32,6 +32,17 @@ describe('Compose', () => {
     expect(await page.checkToRecipientInputFocus()).toBeTruthy();
   });
 
+  it('should update action bar text to subject', async () => {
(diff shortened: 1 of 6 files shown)


tadzik:
Hi @WillEllis, thanks for the extra tests! We can always use more of these :)

As you may have noticed, our CI was not entirely happy with the commit message: we have a standard for these defined in [CONTRIBUTING.md](https://github.com/runbox/runbox7/blob/master/CONTRIBUTING.md#commit-message-guidelines). Yours would look somewhat like `test(e2e): add tests for creating folders <and whatever you desire>`. Beware of not overdoing it with the length.

You'll need to edit your existing commit with something like `git commit --amend`. You can then verify locally if everything is alright with `npm run policy`.

After everything looks good, feel free to `--force` push into this branch so that the PR updates :)

PR #324, opened 2019-10-15
fix(account-app): Implement the missing bits of PaymentRequest handling

tadzik:
This fixes Apple Pay and others.


tadzik:
The diff is best viewed with the whitespace changes disabled: https://github.com/runbox/runbox7/pull/324/files?w=1

PR #325, opened 2019-10-15
feat(calendar): Implement client-side ical parsing

tadzik:
This doesn't yet make the whole thing work, but it passes all the tests.
Time to write more tests :)


tadzik:
Recurring events are currently broken, so it's not ready to merge yet

PR #327, opened 2019-10-17
Hernan604/dkim updates

hernan604:


src/app/dkim/dkim.component.html
```diff
@@ -82,8 +82,9 @@ <h2 class='subheading-2'>{{domain}}</h2>
 
                     <mat-grid-list cols="4" rowHeight="30px" style='' *ngFor="let k of keys">
                         <mat-grid-tile style='border-bottom: 1px dotted #dedede;'><div class="grid_align_left">{{k.selector}}</div></mat-grid-tile>
-                        <mat-grid-tile style='border-bottom: 1px dotted #dedede; border-left: 1px dotted #dedede'><div class="grid_align_left">{{k.is_cname_correct ? 'Yes' : 'No'}}
-                                <a href="javascript:void(0)" *ngIf="!k.is_cname_correct" (click)="check_cname(k)" color="primary">Force re-check {{k.selector}}</a>.
+                        <mat-grid-tile style='border-bottom: 1px dotted #dedede; border-left: 1px dotted #dedede'>
+                            <div class="grid_align_left">
+                                {{k.is_cname_correct ? 'Yes' : 'No'}}<a *ngIf="k.is_cname_hosted_by_runbox && !k.is_cname_correct" href="javascript:void(0)" (click)="check_cname(k)">Check</a>
```
tadzik: is the `href` actually needed when (click) is there already?

src/app/rmmapi/rmmhttpinterceptor.service.ts
```diff
@@ -63,6 +65,10 @@ export class RMMHttpInterceptorService implements HttpInterceptor {
                             req.url === '/ajax_mfa_authenticate' &&
                             r.body.is_2fa_enabled === '1') {
                                 console.log('proceed with 2fa login');
+                        } else if (
+                            req.url !== '/ajax_mfa_authenticate'
+                            && r.body.errors && r.body.errors.length) {
+                            this.snackBar.open(r.body.errors.join(' '), 'Dismiss');
```
tadzik: I'm not sure I'm a fan of this. There are already places in our code that catch request errors by themselves (and often display a snackbar when they find them, like here: https://github.com/runbox/runbox7/blob/master/src/app/contacts-app/contacts.service.ts#L66

Having a global HTTP interceptor catch all errors and pop them up on a screen may cause duplicate messages, or messages popping up in places where it doesn't make sense.

I think this should be handled by the caller of the request itself, not in the interceptor itself.

PR #328, opened 2019-10-17
style(draftdesk): Add a message to an empty draftdesk

tadzik:
It now looks less like a rendering error when no drafts are available :)


tadzik:
@gtandersen I used the same CSS class as the "no messages" screen for visual consistency, though maybe something more visible would be better here.

PR #329, opened 2019-10-23
Canvastable scroll lagging and column resize fixes

petersalomonsen:
when touching the table a focus event is currently triggered to take focus away from the search input field when scrolling the table. On iOS devices this causes lagging scroll, and the fix is as simple as just removing the call to `this.canv.focus()`.

Now this contains no replacement for taking away the focus from the input field when scrolling, so if you still want this you'll have to find another alternative for that.

Also fixed that if you drag the resizer next to the subject column causes the mailviewer to open, and removed the animationframethrottler which only seems to be used by the resizer. If a throttle mechanism is needed (seems not to be), probably better to find another alternative mechanism than requesting animation frames


tadzik:
LGTM too, thanks! :)

PR #334, opened 2019-10-28
feat(webmail): Add an option to hide the message pane

tadzik:
Needs a way to add it back when removed by accident. @gtandersen, any idea for a good place to add the setting?


gtandersen:
Nice work @tadzik.

If this option only affects the vertical preview pane we should perhaps show the option in that pane's menu only?

Minor suggestion: Change texts to "Hide preview pane unless a message is opened" and "Keep preview pane open".

tadzik:
I changed the wording now, and also disabled the preview pane (and the toggle) in mobile view – otherwise it could take over the entire screen and prevent anything from being shown at all.

PR #336, opened 2019-10-30
Calendar overview

tadzik:
This implements a basic (for now) calendar overview screen. I based it on tadzik/jcal to make sure it's using all the latest and greatest and to reduce the possibility of conflicts.

Currently looks somewhat like this:
2019-10-30-144214_1016x790_scrot

Suggestions welcome :)

### DIFF
--- a/src/app/calendar-app/calendar-app.component.html
+++ b/src/app/calendar-app/calendar-app.component.html
@@ -28,22 +28,29 @@
 <ng-template #viewModeButtons>
     <button mat-raised-button
         class="calendarToolbarButton"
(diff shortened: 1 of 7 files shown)


gtandersen:
Looks great @tadzik!

tadzik:
No longer a WIP, but got some more changes in a rebase, so could use another review.

Now features a proper RunboxCalendarView enum instead of abusing `null` as "overview" in the builtin one, remembers the last used view (which was a requested feature even before this), and has an extendable limit to how many upcoming events are being displayed.

PR #340, opened 2019-10-31
Fix wrong folder ordering and layout

tadzik:
This is a followup-fixup for https://github.com/runbox/runbox7/pull/286, which was wrong.

This fixes the problem fixed in that PR without introducing new bugs. Forum users are reporting for the fix to be working: https://community.runbox.com/t/folder-list-gone/531/33


tadzik:
This actually broke sorting completely, as noticed by Travis (!).

A new, better PR is here: https://github.com/runbox/runbox7/pull/341

PR #341, opened 2019-10-31
fix(folderlist): Make sure folders appear in the right order

tadzik:
The previous sorting method could accidentally put child folders above
their parents, which then broke the tree rendering algorithms.

This makes sure that sorting happens at each tree layer separately,
ensuring that the structure is preserved regardless of folder
priorities.

### DIFF
--- a/src/app/folder/folderlist.component.ts
+++ b/src/app/folder/folderlist.component.ts
@@ -101,13 +101,6 @@ export class FolderListComponent {
                 const parentStack: FolderNode[] = [];
                 let previousNode: FolderNode = null;
 
-                // the loop below only handles folders correctly
-                // if it visits parents before it visits their children.
-                // This ensures that it happens
-                folders.sort((a, b) => {
-                    return a.folderLevel - b.folderLevel;
-                });
-
                 folders.forEach((folderCountEntry, ndx) => {
                     const folderNode: FolderNode = { children: [], data: folderCountEntry};
 
@@ -134,6 +127,16 @@ export class FolderListComponent {
 
                     previousNode = folderNode;
                 });
+
+                // Descend into each folder and order its subfolders, but only the direct children.
+                // We need to avoid the situation where sorting would put children above their parents
+                // as that'd mangle the output.
+                const sortSubfolders = (node: FolderNode) => {
+                    node.children.sort((a, b) => a.data.priority - b.data.priority);
+                    node.children.forEach(child => sortSubfolders(child));
+                };
+
+                treedata.forEach(node => sortSubfolders(node));
 
                 return treedata;
             })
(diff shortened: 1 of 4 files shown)


petersalomonsen:
Yes, as you've written in the code comments: "Descend into each folder and order its subfolders, but only the direct children" seems like a good strategy compared to what it is now ( after the previous fix my subfolders ended up as children of the last folder ).

src/app/folder/folderlist.component.ts
```diff
@@ -393,7 +396,16 @@ export class FolderListComponent {
             sourceIndex++;
         }
 
-        // console.log('after rearrange', folders.map(f => `${f.folderId} ${f.folderPath}`));
+        // Adjust the priorities, client-side
+        // Eventually the backend will do it by itself, and we'll se the results on the next reload.
+        // We'd rather see an instant change in the UI though -- so let's calculate temporary priorities
+        // so that everythings looks alright.
+        let priority = 1;
+        for (const f of folders) {
```
petersalomonsen: is this needed? when sent to the server below, the array of folder ids are sent in the order they appear. (priority field is only used for the client side sorting)

tadzik: It's not needed for the server: it's needed for the folderlist component. Few lines down we do [`this.messagelistservice.folderCountSubject.next(folders);`](https://github.com/runbox/runbox7/blob/04beb7e36a41cddc0495fec52fe5eda366a4c2c7/src/app/folder/folderlist.component.ts#L410), which goes straight to the UI [here](https://github.com/runbox/runbox7/blob/04beb7e36a41cddc0495fec52fe5eda366a4c2c7/src/app/folder/folderlist.component.ts#L94) before it goes to the API. The UI then re-renders them, and if this line here is missing, it takes old priorities into account more seriously than the order of entries, so after the drag and drop the UI on the screen remains the same as before: even though the API gets updated, and the next page load will show the folders after the change.

I hoped that the comment explains this, but perhaps it needs to be more detailed :)

tadzik: The reason it wasn't needed before is because the sorting happened in the API service rather than the UI, so when we were feeding the UI with new, after-drag-and-drop data, the re-sorting would not happen like it does now. It's a bit of a hack, but the alternative would be to teach the API service about the tree structure that we need, which feels a bit awkward when it's already doing the flattening of the tree it gets in the first place :­P

petersalomonsen: OK - sounds good :-) Good that you're fixing this !

tadzik: Sweet :) Wanna give me your Seal of Approval? ;)

PR #344, opened 2019-11-01
Make calendar (more) mobile friendly

tadzik:
This includes several refactors to the webmail to allow reusing its good ideas in other apps – this time it's gonna be the calendar :)

The sidemenu now behaves similarly to the one in the webmail: going over the content and becoming closable with the runbox logo. The month-day-view template was adjust to show less information: event details can still be read when clicking on individual days.

Notable missing bit is navigation between sub-apps in mobile view: this can now be done in app-sidenav-menu and thus shared between sub-apps.

Some screenshots:

2019-11-01-175618_956x1053_scrot

2019-11-01-175336_956x1053_scrot

PR #357, opened 2019-11-13
feat(menu): Add Contacts and Calendar to sidenav menu for mobile view…

gtandersen:
… and small screens.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -25,12 +25,11 @@
       <mat-divider></mat-divider>
 
       <div class="sidenavMainMenu">
-        <a mat-button (click)="compose();" matTooltip="Compose">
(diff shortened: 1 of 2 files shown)


tadzik:
I think this is covered by https://github.com/runbox/runbox7/pull/344/commits/2e3be517b2e14b0d75494f58a71f8ec1b00d8a49.

I stole some of your fixes from here into 43e56659e3e7a33782199552ab31c02532df642b, giving us a layout that looks somewhat like this:
5AvurHS6yp

gtandersen:
Looks great @tadzik, especially if we add Contacts! :)

hernan604:
So this should be deleted ?

gtandersen:
Yes!

PR #360, opened 2019-11-15
feat(loginerrors): render more login errors

hernan604:


src/app/login/login.component.ts
```diff
@@ -101,19 +104,57 @@ export class LoginComponent implements OnInit {
...
+        }
+        if (loginresonseobj.user_status > 0 && loginresonseobj.user_status < 5 && loginresonseobj.error) {
+            this.login_error_html = loginresonseobj.error;
+            return true;
+        }
+        return false;
+    }
+
     private handleLoginResponse(loginresonseobj: any) {
         if (loginresonseobj.user_status > 0 && loginresonseobj.user_status < 5) {
```
tadzik: this should probably reuse login_errors_reset() – the else is forgetting to reset login_error_html. It's probably safer to reset it initially even if login_error_html is going to be set later.

src/app/login/login.component.ts
```diff
@@ -80,40 +80,61 @@ export class LoginComponent implements OnInit {
...
+            this.login_error_html = '<p>'+loginresonseobj.error+'</p>';
+        }
+        this.login_error_html = "<p>"+loginresonseobj.message + ': ' + (loginresonseobj.error||'error')+"</p>";
+    }
+
     private handleLoginResponse(loginresonseobj: any) {
         if (loginresonseobj.user_status > 0 && loginresonseobj.user_status < 5) {
-            this.accountexpired = true;
-            this.loginerrormessage = null;
+            this.login_error_html = loginresonseobj.error;
```
tadzik: Not using handleLoginError here?

hernan604: now its using it

PR #376, opened 2019-11-25
feat(update-notifier): Include commit hash in app update data

tadzik:
This way we can display the upcoming version properly in the update
notes (as opposed to the old build timestamp that we used to show).

Also, knowing both the previous and the next commit hash (after this has
been deployed for a while) we'll be able to link users to changelogs
generated on the fly from a given commit range.

### DIFF
--- a/ngsw-config.json
+++ b/ngsw-config.json
@@ -1,7 +1,8 @@
 {
   "index": "/index.html",
   "appData" : {
-    "commit": "__COMMIT_HASH__"
+    "commit": "__COMMIT_HASH__",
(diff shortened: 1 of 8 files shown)


gtandersen:
Good initiative @tadzik -- perhaps we could include the timestamp too though, as it's a useful piece of information?

tadzik:
It'll still be there in the "Runbox 7 build time" field, but we can add it to the update prompt as well.

tadzik:
Looks like this now:
2019-11-26-165440_772x204_scrot

PR #377, opened 2019-11-25
feat(login): include checkbox option to use legacy and stay logged in

hernan604:


hernan604:
https://github.com/runbox/runbox7/pull/378

PR #378, opened 2019-11-25
feat(login): include checkbox option to use legacy and stay logged in

hernan604:


tadzik:
Looking good, but the commit message looks a bit confusing :) Can you make it in line with the changes being made, something like `feat(login): added checkboxes for bla bla bla`?

This will make it easier to spot the changes in the future while reading the log, and make it easier to write marketing release notes from last week's changes :)

PR #383, opened 2019-11-29
feat(account-app): Offer a redirect back to rmm6 if stripe payment form fails to load

tadzik:
@gtandersen I'm not sure about the wording here: I initially included "some browser extensions might be blocking it", which can be true in some cases (I've seen overly aggresive uBlock block the Stripe domain for no good reason), but it also may less technical users feel like we're doing something dishonest and the browser is right to block it. I couldn't think of any way to be helpful while not being misleadingly scary, so I just left it like this.

### DIFF
--- a/src/app/account-app/stripe-payment-dialog.component.html
+++ b/src/app/account-app/stripe-payment-dialog.component.html
@@ -28,6 +28,16 @@
 }
 </style>
 
+<ng-template #legacyFallback>
+    <p> Payment form not loading? </p>
+
+    <p>
+        Make sure your browser extensions are not blocking access to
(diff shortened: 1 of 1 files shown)


gtandersen:
@tadzik Perhaps we could say something like "Please ensure that your browser isn't blocking connections to Stripe" (possibly with the exact domain name) as it might be less concerning when we include a name the user should already be familiar with via our UI?

tadzik:
Changed it to look like this: 
2019-11-29-142738_550x397_scrot

gtandersen:
@tadzik Great! Minor nitpick: The `</a>` should come before the `.`.

PR #384, opened 2019-11-29
fix(calendar): Don't try to import events when user cancelled it

tadzik:
It resulted in tons of PUT queries peformed on /rest/v1/calendar/ics/undefined,
wasting time, bandwidth, and making our error logs look stupid ;)


hernan604:
great!

PR #386, opened 2019-11-29
style(canvastable): Fix overflow and adjustment of column headers

tadzik:
Flex prevents the optionally displayed sorting arrow
from shifting the text downwards.

Overflow does what the overflow does :)

Takes us from this:

2019-11-29-173506_156x32_scrot

To this:

2019-11-29-173522_146x28_scrot

I tried to give it a little bit of padding on the right too, so that it doesn't cut off pixel-perfectly, but my CSS “skills” failed me :/


tadzik:
> Currently the element has a `margin-right: 5px` -- this doesn't work anylonger?

Must be a different element, this one does not. And on this one, `margin-right` doesn't seem to do anything :/

PR #400, opened 2019-12-07
style(menu): Add tooltip to menu items that link to Runbox 6.

gtandersen:


tadzik:
Can we refactor that into something like `rmm6tooltip` in the .ts component so that we can avoid repetition in the template? Then we can do `[matToolTip]="rmm6tooltip"` in the HTML template.

gtandersen:
@tadzik Absolutely.

tadzik:
Needs a rebase on top of current master.

PR #405, opened 2019-12-07
feat(identities): identities

hernan604:


src/app/aliases/aliases.editor.modal.ts
```diff
@@ -0,0 +1,263 @@
...
+     }
+    .header-image {
+       border-radius: 50%;
+       flex-shrink: 0;
+       background-image:url(/_img/avatar.svg);
+       background-size: cover;
+    }
+    `],
+    template: `
+        <mat-card class="mat_card {{css_class()}}" style="padding: 0px">
```
tadzik: `css_class()` may be called at every render frame here (in the worst case: see [here](https://medium.com/showpad-engineering/why-you-should-never-use-function-calls-in-angular-template-expressions-e1a50f9c0496)). 

It would be better to have this set statically on object creation. From what I can see in `profiles/profiles.component.ts`, these `is_whatever` settings are set exactly once on object creation. Since this is a dialog, perhaps it'd be better to pass the "mode" (maybe as an enum, `EditorMode::Create/Update/Delete`) with the `data` and then set the relevant CSS classes just once, in the constructor? Then we don't have to look it up every frame.

src/app/aliases/aliases.editor.modal.ts
```diff
@@ -0,0 +1,263 @@
...
+        private http: HttpClient,
+        public snackBar: MatSnackBar,
+        public dialog_ref: MatDialogRef<AliasesEditorModal>,
+        @Inject(MAT_DIALOG_DATA) 
+        public data: any,
+        public rmm: RMM,
+    ) {
+        this.load_allowed_domains();
+    }
+    css_class() {
```
tadzik: As above: I think an [Enum](https://www.typescriptlang.org/docs/handbook/enums.html) would do nicely here. Its value could even be the string describing the CSS class, then you won't need an additional lookup table for it:

```
enum EditorMode {
    Create = 'create',
    ...
}
```

src/app/aliases/aliases.lister.ts
```diff
@@ -0,0 +1,162 @@
...
+  visible_code_check = {};
+  constructor(public dialog: MatDialog,
+    public snackBar: MatSnackBar,
+    public rmm: RMM,
+  ) {}
+  edit (item): void {
+      item = JSON.parse(JSON.stringify(item))
+      this.dialog_ref = this.dialog.open(AliasesEditorModal, {
+          width: '600px',
+          data: item
```
tadzik: Follow up on the above: here it could be ```{ width: '600px', data: { item: item, mode: EditorMode.Update }```. It is then impossible to forget/fail to set the right mode (as long as the constructor remembers to complain about missing mode :))

src/app/app.module.ts
```diff
@@ -92,6 +97,9 @@ const routes: Routes = [
         component: HeaderToolbarComponent
       },
       { path: 'domainregistration', component: DomainRegisterComponent},
+      { path: 'identities', component: ProfilesComponent},
+      { path: 'dev', component: DevComponent},
```
tadzik: Not sure if we should have this in a production build. It's harmless of course, but feels a bit off. Perhaps we could move it to a separate app, like with have with rmm6 or e2e currently: something that we can build ourselves, but not something we ship to the customers?

src/app/app.module.ts
```diff
@@ -158,12 +166,16 @@ const routes: Routes = [
...
     RunboxWebmailAPI,
+    RMM,
     RMMAuthGuardService,
     StorageService,
     { provide: HTTP_INTERCEPTORS, useClass: RMMHttpInterceptorService, multi: true}
   ],
   bootstrap: [MainContainerComponent],
-  entryComponents: [MoveMessageDialogComponent]
+  entryComponents: [MoveMessageDialogComponent,
+    ProfilesEditorModal,
```
tadzik: Shouldn't this be in the identities module? No other (non webmail related) dialog is specified here in the app.module.ts.

src/app/compose/compose.component.ts
```diff
@@ -59,6 +59,8 @@ export class ComposeComponent implements AfterViewInit, OnDestroy, OnInit {
     public uploadprogress: number = null;
     public uploadingFiles: File[] = null;
     public saved: Date = null;
+    has_pasted_signature;
+    signature;
```
tadzik: Can we give these some types? I assume the latter might be `any` anyway, but maybe we can do better than plain JS here :)

src/app/domainregister/domainregister.component.ts
```diff
@@ -629,7 +629,7 @@ export class DomainRegisterComponent implements AfterViewInit {
   //  .timeout(60000) 
   //  .subscribe(
   //      data => {
-  //          var reply = data.json()
+  //          var reply = data
```
tadzik: Looks weird to have changes in commented out code :P Perhaps this should just be deleted if it's not used?

src/app/app.module.ts
```diff
@@ -92,6 +97,9 @@ const routes: Routes = [
         component: HeaderToolbarComponent
       },
       { path: 'domainregistration', component: DomainRegisterComponent},
+      { path: 'identities', component: ProfilesComponent},
```
tadzik: If we settled on "Identities" then `ProfilesComponent` should probably be renamed.

src/app/profiles/main.ts
```diff
@@ -0,0 +1,6 @@
+import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
+import { ProfilesModule } from './profiles.module';
+import { enableProdMode } from '@angular/core'
+
+const platform = platformBrowserDynamic();
+platform.bootstrapModule(ProfilesModule);
```
tadzik: I don't think I've ever seen a main in our app before :) What does this do?

src/app/profiles/profiles.component.spec.ts
```diff
@@ -0,0 +1,24 @@
+import { TestBed, async } from '@angular/core/testing';
+import { ProfilesComponent } from './profiles-app.component';
+import { ProfilesModule } from './profiles-app.module';
+import { RunboxWebmailAPI } from '../rmmapi/rbwebmail';
+import { RouterTestingModule } from '@angular/router/testing';
+import { of, Observable } from 'rxjs';
+import { Profile } from './profile';
+import { LocationStrategy, APP_BASE_HREF } from '@angular/common';
+
+describe('ProfilesComponent', () => {
```
tadzik: Can we have some basic tests here? :)

src/app/profiles/profiles.component.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  show_error (message, action) {
+    this.snackBar.open(message, action, {
+      duration: 2000,
+    });
+  };
+
+  add_alias (): void {
+      window.location.href = '/mail/account_alias';
```
tadzik: It seems like either the redirect to rmm6 or the AliasesEditor is unnecessary here – I suspect the former, and in that case the href and return should probably go away.

src/app/profiles/profiles.component.ts
```diff
@@ -0,0 +1,134 @@
...
+      });
+  }
+
+  add_profile (type): void {
+    let item = {type: type};
+    this.dialog_ref = this.dialog.open(ProfilesEditorModal, {
+        width: '600px',
+        data: item,
+    });
+    this.dialog_ref.componentInstance.aliases_unique = this.aliases_unique;
```
tadzik: Like the editor mode, I think this should be included in `data` above.

src/app/profiles/profiles.editor.modal.ts
```diff
@@ -0,0 +1,583 @@
...
+            this.localpart = data.profile.email;
+            data.profile.preferred_runbox_domain = this.localpart;
+        }
+    }
+    save() {
+        if ( this.is_create || this.is_create_main ) { this.create() }
+        else { this.update() }
+    }
+    create(){
+        this.is_busy = true;
```
tadzik: Remember to set this to false not only when the HTTP query succeeds, but also if it fails. Right now it's only set back to false if the HTTP query didn't fail, and the REST response is `success`.

src/app/profiles/profiles.editor.modal.ts
```diff
@@ -0,0 +1,583 @@
...
+      });
+    };
+    onchange_field ( field ) {
+        if ( this.field_errors && this.field_errors[field] ) {
+            this.field_errors[field] = [];
+        }
+        if ( field === 'preferred_runbox_domain' ) {
+            const data = this.data;
+            const selected_domain = data.profile.preferred_runbox_domain;
+            ['email'].forEach((attr)=>{
```
tadzik: Is there a need for this `forEach` here if it's just a single string?

src/app/rmm/alias.ts
```diff
@@ -0,0 +1,106 @@
...
+export class Alias {
+    public profiles:any;
+    public aliases: any;
+    public aliases_unique: any;
+    public aliases_counter: any;
+    constructor(
+        private app: RMM,
+    ) {
+    }
+    load() {
```
tadzik: I'm not a fan of having to `load()` things manually after having created an object in order to have the data available. If there's no reason to load them every time, then perhaps it'd be better to load them automatically whenever someone tries to use the data. Something like this:

```
private _profiles: any;
get profiles(): any {
    if (!this.data_loaded) {
        this.load(); // sets this.data_loaded
    }
    return _profiles;
}
```

Right now it's an object that requires extra steps to be used correctly, and otherwise might lead to subtle bugs that won't be easy to track down, because nothing was actually done.

src/app/rmm/me.ts
```diff
@@ -0,0 +1,52 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+// 
+// You should have received a copy of the GNU General Public License
+// along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
+// ---------- END RUNBOX LICENSE ----------
+import { timeout, share } from 'rxjs/operators';
+import { RMM } from '../rmm';
+
+export class Me {
```
tadzik: This goes for all the new RMM classes (which are a good idea: the one giant rmmapi has gotten a bit silly by now...), but I think it'd be good to make them `@Injectable` so that every component using them doesn't need to create them by itself (which also makes testing near impossible).

Right now every component using `RMM::Me` has to created it by itself and then load the values (which I'd also rather have eliminated, as described in the comment above). That means that we're performing a query to `/rest/v1/me` for every component using `RMM::Me`, which is pretty wasteful.

Making this `@Injectable` means that any object using it can just declare a depedency on this in its constructor arguments (like we do now with rmmapi) and then only one copy is being created in the entire app, and no unnecessary queries are performed. Also, it becomes possible to replace them with stubs and mocks in tests.

src/app/rmm/useragent.ts
```diff
@@ -0,0 +1,29 @@
...
+// General Public License for more details.
+// 
+// You should have received a copy of the GNU General Public License
+// along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
+// ---------- END RUNBOX LICENSE ----------
+import { timeout } from 'rxjs/operators';
+import { HttpClient, HttpHeaders } from '@angular/common/http'
+import { RMM } from '../rmm';
+//ABSTRACT: this class has the http client that is used on this app
+export class UserAgent {
```
tadzik: I'm not sure this needs to exist: why not just have an `HttpClient` in the `RMM` class?

src/app/profiles/profiles.component.ts
hernan604: aliases will be added from rmm6

src/app/profiles/profiles.component.ts
hernan604: Not really.

The idea here is that 'data' is the object received from the backend.

This object will populate the form which is inside the modal.

When users change the form, its reflected in 'data' which can be used to post back to the server.

The modal can have its own properties also. But they are not in data... they are properties for the modal. And the angular modal allows this.dialog_ref.componentInstance[$attr] to update them from outside (the component creating the modal in this case).

src/app/rmm/alias.ts
hernan604: when you add this hack, leave load there so we can re-load it

src/app/rmm/me.ts
hernan604: definately should not call multiple times... the idea is that 'rmm' is 1 instance.

So this.rmm.aliases.load() will probably consume aliases endpoint. and store data this.rmm.aliases.data so other classes can use this.rmm.aliases.data without reloading it.

hernan604: Also,

Dont add @Injectable in RMM Class..

First im trying to avoid all that @*unecessary-stuff*

The RMM class is injectable, and 'this.rmm' is a normal OO class that has atributes, instances, normal stuff.

If you see the Me load() method, it has a 

`
     load() {
         if ( this.data ) { return; }
`

So 3 places in profiles call load() at the same time, and neither of the calls were terminated. So it calls 3 times. If data were to have a value, it would not call the endpoint again.

This class has an attribute which is this.is_busy which is true when its fetching values. To fix the multiple calls it possible to alter the code to:
`
     load() {
         if ( this.data || this.is_busy ) { return; }
`

That will make the endpoint be called only 1 time.

We can of course change how this works but lets keep it simple OO.

src/app/rmm/useragent.ts
hernan604: Since the RMM class is simple OO, the useragent could be an instance of some other HttpClient. For example axios. This way other frameworks are able to re-use this same RMM class and just create an instance with their own UserAgent.. no need to have any angular deps to use the RMM class in an react project.

hernan604: The only class that ties RMM to angular is probably the useragent class which uses:

import { HttpClient, HttpHeaders } from '@angular/common/http'

The more RMM class is framework agnostic, the best i think...

src/app/rmm/me.ts
tadzik: Imho `@Injectable` *is* simple OO. It's just the readymade implementation of one of the SOLID rules, which is arguably basic OO :)

I see the point about the single instance: I didn't notice that Me etc are all part of RMM and you're supposed to use it as a whole. In that case you will indeed have only one copy of everything.

But it does feel, currently, like a bit of a downgrade from RunboxMe: it requires me to pull in the entire God-object as a dependency effectively, and an object that does its own UI for error handling. I like the idea of keeping things simple and synchronous as much as possible, since it does make everything easier, but as it stands I think things are made a bit *too* simple. If you want to show some data in the template then cool: it works for that, if you just bind to `Me.data`. But consider `StorageService` for instance: it needs to know part of the Me, the uid, to implement user-specific localStorage. There's no template, nothing to bind to. It needs a way to say "give me the UID, once it becomes available, and then lookup the localStorage with the key mashed together with the UID". The `RMM.Me` class doesn't make it possible to implement that. Grabbing `.data` is not good enough: it may be not there yet. I can call `.load()` and subscribe on the result... but if someone called .load before me then I'm either duplicating calls (currently), or if we apply the `if ( this.data || this.is_busy ) { return }` fix you mentioned, I'm not going to get the HTTP query object and will have nothing to subscribe to.

We keep can keep as it is, sure. But I don't see myself using it at this point just yet :)

src/app/rmm/useragent.ts
tadzik: We're not being RMM agnostic here though: we're just putting our own box over Angular, and keeping its interface. If we were to move to axios tomorrow we'll still need to rewrite the entire RMM class at the very least, and since `load()` and some other methods (`create()`, `resend()` and `delete()` in Profile, and probably a lot more) “leak out” the underlying request object, in the end or entire codebase is still tightly coupled with Angular's HttpClient specifically, and not on any abstraction. And if you did actually build a different HTTP client that uses the same interface as HttpClient, you can swap it for the entire project just as well: since we're using Dependency Injection everywhere, we can provide anything we want as the HttpClient, [like we currently do in our unit tests](https://github.com/runbox/runbox7/blob/master/src/app/mailviewer/singlemailviewer.component.spec.ts#L79). That's the purpose of it being Injectable instead of something you `new` yourself: so that you can trivially substitute something else that implements the same interface.

src/app/app.module.ts
hernan604: Tried to remove this and added in profiles.modue.ts however the modal does not open.
I think it needs to be here.

src/app/app.module.ts
hernan604: Maintain another app just for this? i rather have it available to check easily... and also simple to link from if necessary.

src/app/rmm/me.ts
hernan604: I did not consider any localstorage for that RMM class.. but the data is in this.rmm.me.data so it could be easily jsonstringified or stored in localstorage when its fetched, or processed for xapian when its fetched.

If you think its too simple, that is good, that was the point :)

Any of it can be changed, upgraded, etc.. im just trying to create this 'model' so its easy to create screens that list data and allow to edit them via simple crud.

When this version is better you may use it

src/app/app.module.ts
tadzik: That's because your ProfilesModule is not actually being loaded into the app module. I have no idea how or why it even works in the first place: it did not load for me in the `npm run start` mode because of the unloaded module. I fixed it [here](https://github.com/runbox/runbox7/commit/00a2fdfdfbf1de875303548375fd5c6de4659778) – now it also passes the tests :)

src/app/rmm/me.ts
tadzik: It's not about localstorage itself, it's about the asynchronous nature of the data :) Your approach works fine if the async data from your components is put straight into the Angular template, but is unfit for cases if you need some extra processing done to it.

> If you think its too simple, that is good, that was the point :)

Aha, I think I see your approach now :) It definitely does work well for simple cases, with inbuilt error handling and all that – makes sense. For the heavy lifting we can still use the async stuff where necessary.

tadzik: One nitpick that I have though: can we get some types added into this and make this consistent?

`load()` sometimes returns the async query that it's running, which will resolve to the data later, but sometimes it does not. It may lead to a situation where a code works one day, because it was the first component to call .load(), and a week later some other component gets to it first and the old one will end up doing `undefined.subscribe` because of reasons beyond its control.

I think it'd be safer to have the type declared in Typescript: make it `load(): void { }` or `load(): Observable<any> { }`, which ever you see more fitting (the former is probably easier, the latter may be more useful), and this will protect us from accidentally misusing it.

hernan604: if undefined.subscribe situation exists this will result in error and thats not wanted.

I will use the Observable<any> suggestion

tadzik: You can probably use `Observable.of` (I'm not sure if that's what it's called) to return something compatible in case the data is already there (the early exit from `load()`)

src/app/rmm/me.ts
```diff
@@ -27,8 +28,8 @@ export class Me {
         public app: RMM,
     ) {
     }
-    load() {
-        if ( this.data ) { return; }
+    load(): Observable<any> {
+        if ( this.data || this.is_busy ) { return; }
```
tadzik: Isn't this bare `return;` violating the type constraint?

hernan604: what should it return

src/app/aliases/aliases.editor.modal.ts
hernan604: resolved

src/app/aliases/aliases.editor.modal.ts
hernan604: resolved

src/app/aliases/aliases.lister.ts
hernan604: resolved

src/app/profiles/profiles.component.spec.ts
hernan604: no tests for now in identities components :~

src/app/app.module.ts
hernan604: Its from identity profile =p 
it was initially profile until identity seemed better

src/app/profiles/profiles.editor.modal.ts
hernan604: because it had the 'reply_to' field

src/app/rmm/useragent.ts
hernan604: There is always room for improvement.

It would be cool to have that agnostic however this is not that important as long as it works inside runbox7 project

PR #411, opened 2019-12-18
style(payment): Make layout more responsive, add button color, and cr…

gtandersen:
…edit card logos.


tadzik:
Rebasing on top of current master should fix the e2e failures (just in case it actually broke anything :))

PR #415, opened 2019-12-28
Hernan604/identities issues

hernan604:


src/app/rmmapi/from_address.ts
```diff
@@ -0,0 +1,56 @@
...
+// Runbox 7 is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+// 
+// You should have received a copy of the GNU General Public License
+// along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
+// ---------- END RUNBOX LICENSE ----------
+
+export class FromAddress {
```
tadzik: Good to see this refactored out :)

src/app/rmm/plugin/tinymce.plugin.ts
```diff
@@ -0,0 +1,68 @@
...
+// ---------- END RUNBOX LICENSE ----------
+import { timeout, share } from 'rxjs/operators';
+import { Observable } from 'rxjs';
+import { RMM } from '../../rmm';
+
+declare const tinymce: any;
+
+export class TinyMCEPlugin {
+    constructor(
+        public app: RMM,
```
tadzik: This looks unused here – and I assume it is, since there's no reason why the editor should need an RMMapi :­P

Since this is copypasted from compose.component.ts, can you make ComposeComponent use TinyMCEPlugin so that we don't have duplicate code?

src/app/rmm.ts
```diff
@@ -36,6 +37,8 @@ export class RMM {
     public alias: Alias;
     public me: Me;
     public runbox_domain: RunboxDomain;
+    public tinymce_plugin: TinyMCEPlugin;
+    public plugins: any = {};
```
tadzik: I don't think this needs to exist if we only have one plugin. https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

Also, what does the editor have to do with the API? This is just mixing concerns and creating a God-object like Webmail.pm – I thought we wanted to move away from that, otherwise we could've just sticked to existing rmmapi.

hernan604: Its like why there are powerplugs availables in everyhome if some of them are empty and not sold ? The constructors should let the owners do the powerplug themselfs instead of foreseen they will need some eventually ?

tadzik: I don't think this analogy holds. Powerplugs are hard to add later. Code is not. This is like adding an elevator to a single floor house because you may need more later.

tadzik: But more importantly, again: what does an editor do in an API object? Let's not create another Webmail.pm here.

hernan604: fixed

src/app/rmm/plugin/tinymce.plugin.ts
hernan604: fixed

tadzik: Cool, thanks.

tadzik: Perfect! Can we now have the same code reused in ComposeComponent so we don't duplicate things?

tadzik: Wonderful :)

PR #422, opened 2020-01-09
WIP fix(calendar): Fix importing events with a METHOD property

tadzik:
As noted in the comment: once we have the pieces required to do that,
we should (offer to) notify the organizer of the event that we accept
the invitation. For now though, we can at least import them into the
calendar, since a lot of the people are doing that and expect it to
Just Work[tm].


tadzik:
Note: this doesn't actually enable accepting event invitations: it imports the events, but the notifications aren't being sent (yet). Further digging is needed for that :)

PR #435, opened 2020-01-21
Payment special offer

gtandersen:


gtandersen:
This needs to be fed a variable `future_date` which should be `users.user_created` + 24 hours; see L72 in account-upgrades.component.html.

It should then look like this (with a correct timer):

Payment special offer screenshot

src/app/account-app/account-upgrades.component.html
```diff
@@ -1,59 +1,119 @@
...
+      Here you can subscribe to Runbox, or change an existing subscription plan according to your needs.
     <p>
-      Note: When selecting a subscription plan, please note that your current email disk usage is
+      Note: When selecting a subscription plan for your main account, please note that your current email disk usage is
       {{ (me.disk_used / 1024 / 1024 / 1024) | number:'1.2-2' }} GB.
     </p>
 
-    <p>
-        And don't forget to check out
+    <div *ngIf="!(future_date < now)" class="limitedTimeOffer">
```
tadzik: Both `future_date` and `now` don't exist as variables, turning this into `!(undefined < undefined)` (`true`) and thus making it always appear.

src/app/account-app/runbox-timer.ts
```diff
@@ -0,0 +1,137 @@
+// --------- BEGIN RUNBOX LICENSE ---------
+// Copyright (C) 2016-2018 Runbox Solutions AS (runbox.com).
```
tadzik: Should be 2020 now I suppose :) I'm no expert on legal matters, but should be strive to keep this up-to-date? Perhaps a test (with an updater) should be introduced, that makes sure our copyrights are up-to-date?

src/app/account-app/runbox-timer.ts
```diff
@@ -0,0 +1,137 @@
...
+  Output,
+  EventEmitter,
+  NgZone,
+  ViewChild,
+  AfterViewInit,
+  ContentChild,
+  ElementRef,
+  TemplateRef,
+} from '@angular/core';
+import { CommonModule } from '@angular/common';
```
tadzik: Importing modules doesn't do anything in components -- all of these can go :)

tadzik:
> I found one actual bug in there

(Now that I have the link, the aforementioned bug is [here](https://github.com/runbox/runbox7/pull/435#discussion_r369649003))

tadzik:
Looking very nice, good work :)

PR #457, opened 2020-02-05
feat(app): Add an indicator if the API connection is lost

tadzik:


tadzik:
Rebased on top of current master and added an explanatory text the first time it happens (in this session).

PR #466, opened 2020-02-13
Add a way to export contacts

tadzik:
Fixes GH-145


tadzik:
I put it under Settings as it didn't seem significant enough to give it its own sidemenu button – then again, it's not really a setting of any sort. Any better ideas? :)

gtandersen:
I think Settings is a good place for now. :)

Issue #352, opened 2019-11-12
Removing one of two (or more?) identical products from the cart removes them all.


davidbowdley:
I accidentally added two products that were the same to the cart. When I went  tothe cart page and removed one of them it removed them both. This only seems to happen if they are the same product (from the limited testing done).

tadzik:
This is now fixed and will be live in the next release.

PR #468, opened 2020-02-14
Fix adding/removing duplicate products to/from the cart

tadzik:
Fixes GH-352


tadzik:
Can I get some eyes on this? :)

PR #482, opened 2020-03-03
fix(identity): list compose "From" sorted in the order: main, others …

hernan604:
…and aliases


tadzik:
Looks good: but maybe squash the commits so that we end up with one clean changelog entry? :)

PR #494, opened 2020-03-19
WIP Handle recurring event exceptions correctly

tadzik:
WIP as it still needs the ability to create such exceptions.


tadzik:
I'll lay this off for now since so many other things are much more immediate, but a few notes before I forget completely:

The PR is currently in a half-working state, where some odd bugs prevent it from passing the tests that were written for it (and that did pass at some point). It has, however, exposed a design flaw in a calendar that will ultimately prevent it from working for all cases (or it'll at least make it very hard):

The way it's currently written, one RunboxCalendarEvent corresponds to one event on the screen (and implements angular-calendar's CalendarEvent, which has the same purpose). This was a fine assumption that served us well for singular events, but it proved unfit when recurring events came in to play. The bandaid solution was to clone RunboxCalendarEvents for each recurrence, altering their dates for each instance. A `parent` field was introduced, pointing to the "main" event from which clones were made, which allowed editing the "real" event regardless of which clone was being edited.

Another false assumption is that each RunboxCalendarEvent has its own ICAL.Component (`BEGIN/END:VCALENDAR` in .ics), and that each of these ICAL.Component will have exactly one ICAL.Event (`BEGIN/END:VEVENT`) in it. This is false for recurring events, which, when special-cased will have multiple VEVENTs in them (1 or more), and each of them may have a different date, name etc.

With this new understanding a think we'll need a bet of a rewrite. We'll need to implement two separate types: one corresponding (roughly) to a single VEVENT, and that will be the one shown on screen and implementing angular-calendar's `CalendarEvent`, and another that'll be representing a VCALENDAR overlooking individual VEVENTs, specialcases of these etc.

gtandersen:
Nice digging @tadzik and I agree with the conclusion.

tadzik:
Done correctly in https://github.com/runbox/runbox7/pull/790

Issue #512, opened 2020-04-06
Unread message counters are unreliable


tadzik:
The counters for unread messages on the folder list are inconsistent with the message list – and reality :) In the best case they are severely delayed – in the worst, they're completely wrong. I'll to explain why (I think) this happens, and what we can hopefully do to make it better.

The reason why this happens is because they reflect the numbers returned by `https://runbox.com/rest/v1/email_folder/list`. That API call gives us the structure of the folder tree, but also the basic message counters tracked by IMAP (among other things?) and updated every now and then. The good news is that these are then mostly correct whenever a new message arrives and the account is reindexed (or so I think).

Problems start when the user marks the message as read. This triggers the API call to mark the message as read on the server [here](https://github.com/runbox/runbox7/blob/master/src/app/mailviewer/singlemailviewer.component.ts#L380), which triggers a [chain](https://github.com/runbox/runbox7/blob/master/src/app/rmmapi/rbwebmail.ts#L396) of [events](https://github.com/runbox/runbox7/blob/master/src/app/rmmapi/messagelist.service.ts#L62) that ends up with the folder list being refreshed [here](https://github.com/runbox/runbox7/blob/master/src/app/rmmapi/messagelist.service.ts#L71).

If I'm reading this correctly (and my reading comprehension can be so-so when it comes to weird rxjs chains) this may already cause a race condition where we refresh folder counts before the message is successfully marked as read – but that's a minor thing in the grand scheme of things.

The thing worth noting is that our folder counters are *always* server-based and updated and stored in the same place the folder *structure* is. This becomes much more problematic when we use the local index for emails. While the local index doesn't contain the folder structure, it does contain the message flags, including the Seen flag. What the user sees as read/unread messages on the message list comes from here, and this is what can most easily diverge from what the folder list chooses: since the index may (and probably will) update much quicker than an API call that reads the folder data from the server database.

I think that the message list should be authoritative in terms of the folder counters: knowing that there are unread messages on the folderlist when there aren't any in the message list is useless knowledge anyway. It's easier said than done in case of online index, but with a local index it sounds relatively straightforward, and I think that's where we should focus on putting this.

[...]

PR #513, opened 2020-04-06
Fast and correct folder counters

tadzik:
Implements changes described/planned in https://github.com/runbox/runbox7/issues/512.

Suggestions (esp for high-level design) are much welcome.


petersalomonsen:
Yes I think it's the right approach to calculate the folder counts from the local index. I think I even wanted to do this when I worked on this, so very good that you are picking this up :-) It's for sure confusing when the index and the message db is not in sync, and much better if the index can be the master. Server index and local index should probably be more in sync if I remember correct.

gtandersen:
Agree that the folder counts should be built from the index/list currently in use for consistency in the UI -- if it's a local index then use that, while if it's the database then that should be used.

tadzik:
The folder count updates are now instant and consistent with the local index (if it exists). Calculation seems to take around 5ms per folder on desktop firefox – while the queries themselves are quite inefficient, it seems like they're performant enough in practice to not worry about it too much for now :)

tadzik:
This fixes GH-512

PR #514, opened 2020-04-07
feat(multiple_msg_unread): replace endpoint that marks multiple messa…

hernan604:
…ges as unread/unflag to update these attributes with 1 request only

### DIFF
--- a/src/app/rmm/email.ts
+++ b/src/app/rmm/email.ts
@@ -1,0 +1,44 @@
+// --------- BEGIN RUNBOX LICENSE ---------
+// Copyright (C) 2016-2018 Runbox Solutions AS (runbox.com).
+// 
+// This file is part of Runbox 7.
+// 
+// Runbox 7 is free software: You can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the
+// Free Software Foundation, either version 3 of the License, or (at your
+// option) any later version.
+// 
+// Runbox 7 is distributed in the hope that it will be useful, but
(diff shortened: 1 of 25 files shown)


src/app/rmm/email.ts
```diff
@@ -0,0 +1,44 @@
...
+        req.subscribe(
+          data => {
+            const reply = data;
+            if ( reply['status'] === 'error' ) {
+              this.app.show_error( reply['error'].join( '' ), 'Dismiss' );
+              return;
+            }
+          },
+          error => {
+            return this.app.show_error('Could not load aliases.', 'Dismiss');
```
tadzik, line 39: Wrong error message.

e2e/mockserver/mockserver.ts
```diff
@@ -144,6 +144,11 @@ export class MockServer {
...
+                return;
+            }
```
tadzik, line 151: This looks like a leftover from testing and it makes your tests fail (because of double-encoded JSON).

tadzik:
I think the new code doesn't notify the [messageFlagChangeSubject](https://github.com/runbox/runbox7/blob/master/src/app/rmmapi/rbwebmail.ts#L150) (which the messagelistservice observed), so any seen flag updates will only show on the screen after the next index update, possibly making it seem like things aren't working at all.

e2e/mockserver/mockserver.ts
hernan604: removed

PR #527, opened 2020-04-15
fix(compose): Validate TO/CC/BCC fields on save/send

### LINKED ISSUE #39: Validation of recipient addresses
Don't allow empty or invalid recipient field input.

Note that pasted addresses don't seem to be submitted. E.g., copy and paste test@runbox.com into a recipient field and send the message. It will bounce with "no recipient address".

castaway:
Prevents sending or saving draft emails when the TO field is empty or
contains invalid emails. (the latter should already be prevented by
EmailValidator). Event "change" was renamed to "updateRecipient" to
not clash with existing events and be less confusing.

Fixes #39


src/app/compose/emailvalidator.ts
```diff
@@ -35,3 +35,14 @@ export function isValidEmail(email: string) {
...
+/**
+ * Validate each email in the list, return true if all are valid
+ * @param emailList
+ */
+export function isValidEmailList(emailList: string) {
+  if (!emailList) {
+    return false;
+  }
+  return emailList.split(',').every((recipient => isValidEmail(recipient)));
+}
```
tadzik, line 48: This seems to be using 2 spaces for indentation while the rest of the file uses 4 :)

castaway:
Ahh grr. I did change that in my editor settings at some point, probably after this edit..

PR #535, opened 2020-04-17
Switch unit tests to firefox

tadzik:
This eliminates a chromium dependency replacing it with firefox, which should be stock on any linux distro, hopefully including Travis – which will hopefully speed up test runs.

Relying on chromium seems to be causing problems for Cypress on Ubuntu (https://github.com/cypress-io/cypress/issues/7020) which makes development super annoying currently.

I admit to not measuring it, but running unit tests on firefox-headless *feels* faster than it was on chromium too.

Opinions?


tadzik:
Perhaps not a very reliable "benchmark", but we seem to have saved around 30 seconds per job in Travis: [21:44 on Chromium](https://travis-ci.com/github/runbox/runbox7/builds/160746855), [19:20 on Firefox](https://travis-ci.com/github/runbox/runbox7/builds/160748414).

castaway:
Hmm, no firefox (available for install) on the box I'm using, it is a few Debians out of date though.I wonder if an AppImage will run on a headless machine?

tadzik:
> Hmm, no firefox (available for install) on the box I'm using, it is a few Debians out of date though.I wonder if an AppImage will run on a headless machine?

I'll open a ticket about making the browser used configurable with an env var or something: theoretically it should be possible to use electron and not depend on any desktop browser being installed.

Issue #528, opened 2020-04-15
Can get stuck in empty "compose" window


castaway:
1) Start with no drafts
2) Compose a message, save as draft or wait for it to auto-save
3) While in same window, hit the trashbin / delete draft icon
4) Now you are still on the /compose?new page, and cannot "compose" a new message

I think after using the delete draft icon from the /compose?new page, the user should be sent back to plain /compose (or somewhere else useful)

gtandersen:
@castaway Agreed!

castaway:
Looking into this, while I'm poking it - would it also be logical that if you view a contact, and hit the "email this contact" button, and exit the draft, you should be returned to the contact page?

Is there anywhere else (other than reply / forward) that a new draft can be started from?

gtandersen:
Yes, it makes sense to return the user to the screen they originated from.

I don't think there are other ways for a new draft to be started than those mentioned.

castaway:
Merged.

PR #553, opened 2020-04-22
fix(compose): Return user to origin when closing/trashing new drafts

castaway:
Logic for closing / trashing drafts is now:
If its was a new draft (not yet saved when we open it)
* If we started from "Compose", return to draftdesk (/compose)
* If we started from a "reply", return to inbox (/)
* If we started from a "email contact", return to (/contacts)
** This would be better as "the contact", need to figure out if we can do that
* Otherwise: return to drafts desk

### DIFF
--- a/e2e/cypress/integration/compose.ts
+++ b/e2e/cypress/integration/compose.ts
@@ -7,8 +7,8 @@ describe('Composing emails', () => {
     }
 
     function composeNew() {
-	cy.visit('/compose?new=true');
-	cy.closeWelcomeDialog();
+        cy.visit('/compose?new=true');
+        cy.closeWelcomeDialog();
     }
 
     it('should display draft card', () => {
@@ -58,12 +58,13 @@ describe('Composing emails', () => {
     });
 
     it('closing a newly composed email should return to draftdesk', () => {
-    	composeNew();
-	cy.get('button[mattooltip="Close draft"').click();
-	cy.location().should((loc) => {
-	    expect(loc.pathname).to.eq('/compose');
-	    expect(loc.search).to.eq('');
-	});
+        composeNew();
+        
+        cy.get('button[mattooltip="Close draft"').click();
+        cy.location().should((loc) => {
+            expect(loc.pathname).to.eq('/compose');
+            expect(loc.search).to.eq('');
+        });
     });
 
     it('closing a new reply should return to inbox', () => {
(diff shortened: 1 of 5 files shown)


castaway:
Fixes #528

e2e/cypress/integration/compose.ts
```diff
@@ -7,8 +7,8 @@ describe('Composing emails', () => {
     }
 
     function composeNew() {
-        cy.visit('/compose?new=true');
-        cy.closeWelcomeDialog();
+	cy.visit('/compose?new=true');
```
tadzik, line 10: Can we keep 4 spaces here (and in the rest of the file)? :)

src/app/compose/compose.component.ts
```diff
@@ -71,6 +71,8 @@ export class ComposeComponent implements AfterViewInit, OnDestroy, OnInit {
     saveErrorMessage: string;
 
     @Input() shouldExitToTable = false;
+    @Input() shouldExitToDrafts = false;
```
tadzik, line 74: It seems like the `@Input()`s are unnecessary here – also for the pre-existing `shouldExitToTable`.

`@Input()` only matters if you want to set it in the template, like `<compose shouldExitToTable="true">`. We don't seem to be doing that (and the quick glance at `git log` indicates that we never did :P): `src/app/compose/draftdesk.component.html` only uses the `model` `@Input`. These are just regular properties of this class, so the `@Input` can go.

I also wonder if maybe it'd simplify things if we just replaced this set of booleans with something like `exitTo: string`? It would cut the amount of properties/methods we need. Food for thought :)

Come to think of it, while Angular's `Router` (surprisingly) doesn't have anything like a "back" method, apparently there's a Location service that we could use here: https://stackoverflow.com/questions/35446955/how-to-go-back-last-page. Perhaps we could just use that (in some cases)? The added benefit would be being able to go back specifically to the contact that summoned us, not just the `/contacts` list. Another bit to think about :)

e2e/cypress/integration/compose.ts
castaway: hm, while I cant see where it changed back, my checkout of the branch is consistently 4..

src/app/compose/compose.component.ts
castaway: I did wonder if it would be saner to change all "from" places to send a url/route to return to.. I'll look at the Location thingy.

PR #555, opened 2020-04-23
build(tests): Allow Electron and Chrome(ium) in unit tests

tadzik:
This partially reverts 161b66f5dec0740ec11bf5d7f2b63e862ac86334.

Firefox is still the default browser in run-ci-tests, but
Chrome/Electron can be used when launching `ng test` directly.


tadzik:
Not sure why I brought it up back then and what was it good for, so I'll close it now since it obviously didn't prove to be very important after all.

Issue #532, opened 2020-04-16
Add option in individual folder menus to create a new sub-folder


davidbowdley:
Although creating a new sub-folder is possible by clicking on a folder and using the + icon at the top of the folder list it would be useful and more ergonomic to also have an option in the folder menu where you can rename/delete folders.

vertigo220:
To clarify, the add folder pop-up that shows when clicking the plus button should show the folder tree and allow the user to select where in the tree the new folder should be placed, i.e. under the selected branch.

kvviecien:
Fixed in #565

PR #565, opened 2020-05-05
feat(folders): Create sub-folders via individual folder menus

kvviecien:
Add option to create a new sub-folder through individual folders' drop-down menu rather than using the + icon on top after selection of desired folder.

Issue #532


src/app/folder/folderlist.component.html
```diff
@@ -50,6 +50,10 @@
                 <mat-icon color="primary">border_color</mat-icon>
                 <span>Rename</span>
               </button>
+              <button mat-menu-item (click)="addSubFolder(node)">
+                <mat-icon color="primary">create_new_folder</mat-icon>
+                <span>New Folder</span>
```
tadzik, line 55: Perhaps we can make that “New Subfolder” to make it distinct? @geir, @davidbowdley?

The resulting popup still says “New folder”, but also “under XXX”

gtandersen: Nice work @kvviecien, and agree @tadzik -- "New Subfolder" would be technically correct while confirming to the user that they are not creating a folder at the root level.

kvviecien:
Since creating new folders is not covered by Jasmine, I'm working on some tests for it. 
Drafting the PR for now.

Issue #568, opened 2020-05-08
Default From for replies is not trying hard enough


tadzik:
[The code that picks the address to reply with](https://github.com/runbox/runbox7/blob/master/src/app/compose/draftdesk.service.ts#L102) works for simple cases, but could use some improvement. For example, if one of my addresses is `tadzik@github.com`, and I reply to an email sent to `TADZIK FROM GITHUB <TADZIK@GITHUB.COM>`, my default From will be picked instead of the `@github.com` one.

The existing comparison (`===`) in the code linked above needs to be replaced with something more lenient: for example, extracting whatever's in the angle brackets and then matching it while ignoring the casing.

PR #570, opened 2020-05-11
fix(compose): Correct default "From" email for replies

kvviecien:
The Reply option wasn't picking the correct "From" address if the "To"
email was capitalized by the original sender.
The function comparing addresses to find the correct one to reply with
now formats the address, so the casing doesn't break anything.

Fixes: #568


tadzik:
Will it still work if the To puts the address in angle brackets, like `Test mail person <testmail@testmail.com>`? I'm not sure if `fromObj.email` contains only the actual address or if it includes the name component too.

kvviecien:
fromObj has separate properties for name and email, so fromObj.email has just the address in it.
Turns out the alias names actually don't matter, they can be completely different from the email address.

tadzik:
Hm, you're right – I checked again and now it indeed works as advertised. I must've checked the wrong branch (or browser tab...) the first time around ;) Good job!

A test may be a good idea nonetheless – even if it's not actually an edge case after all.

kvviecien:
Sorry for the mess up with changes above. I forgot I'm working on a fork and I didn't update my copy of the repository with recent changes...

PR #586, opened 2020-05-18
Welcome Desk

gtandersen:
Welcome Desk displaying common tasks for new users.


src/app/welcome/welcomedesk.component.html
```diff
@@ -0,0 +1,72 @@
...
+  </div>
+  
+  <div>
+    <h4><mat-icon>mail</mat-icon>Import your email</h4>
+    <a href="/mail/fetch" target="rmm6">Consolidate email</a> from your other email accounts.
+  </div>
+  
+  <div>
+    <h4><mat-icon>people</mat-icon>Import your contacts</h4>
+    Import all your <a href="/contacts">contacts</a> from other email services.
```
tadzik, line 39: Perhaps this should be a `routerLink`, so that it doesn't trigger a full page reload?

src/app/welcome/welcomedesk.component.html
```diff
@@ -0,0 +1,72 @@
...
+  </div>
+  
+  <div> 
+    <h4><mat-icon>domain</mat-icon>Set up Email and Domain Hosting</h4>
+    <a href="/domainregistration" target="rmm6">Register a new domain</a>, or complete <a href="/mail/account_domain">email hosting</a> setup.
```
tadzik, line 44: `/domainregistration` is in runbox7, so this could be a routerLink too (and probably doesn't need/want an rmm6 target).

src/app/welcome/welcomedesk.component.html
```diff
@@ -0,0 +1,72 @@
...
+  </div>
+
+  <div>
+    <h4><mat-icon>devices</mat-icon>Use an email program</h4>
+    Find out how to use your Runbox account with an <a href="https://help.runbox.com/email-programs/">email program</a> or app.
+  </div>
+
+  <div>
+    <h4><mat-icon>card_membership</mat-icon>Subscribe or upgrade</h4>
+    <a href="/account/components">Take full advantage of Runbox</a> with extra accounts and other add-ons.
```
tadzik, line 64: Another potential `routerLink` :)

tadzik:
Also: are we planning to make it show up by default to new users, similar to how the "Welcome to runbox7" dialog shows up now asking to enable local index?

gtandersen:
Thanks @tadzik -- now it's more responsive and with correct links.

And yes, this should be shown when entering Runbox 7 for the first time upon signing up -- but thereafter only on request. This might require rewriting the main app component?

Screen Shot 2020-05-22 at 00 01 55
Screen Shot 2020-05-22 at 00 02 04
Screen Shot 2020-05-22 at 00 01 31
Screen Shot 2020-05-22 at 00 01 44

src/app/welcome/welcomedesk.component.html
gtandersen: Done in 4292cf1.

src/app/welcome/welcomedesk.component.html
gtandersen: Done in 4292cf1.

src/app/welcome/welcomedesk.component.html
gtandersen: Done in 4292cf1.

tadzik:
I was not able to get a nice evenly distributed grid like on the screenshots:

2020-05-26-173619_1919x440_scrot

But it it responsive :)

Maybe we can show this screen to users right after they go through the initial "Welcome to Runbox7" dialog?

Issue #567, opened 2020-05-06
Message with no plaintext body can't be replied to unless HTML is enabled


tadzik:
It's not possible to reply to a message that only contains HTML body: unless HTML mode is actually enabled at that time.

If it's not, it crashes with

```
 TypeError: "t.rawtext is undefined"
    reply https://runbox.com/app/main.c2e2856cd85b02a2f25b.js:1
    ...
```

and provides no other feedback.

PR #588, opened 2020-05-19
fix(compose): Make it possible to reply to emails with no plaintext

kvviecien:
When HTML mode is disabled and there is no plaintext in the email body, it picks an empty string instead for the reply. It should prevent errors from occurring.

Fixes: #567

---

Unfortunately I couldn't properly reproduce the error, but the little fix I did should take care of the problem. 
Additionally, with @tadzik 's suggestion, I changed styling of the in-message notification to make it slightly more visible for the user that they should enable HTML if they want to read the email.


tadzik:
Visibility is much better now:

2020-05-20-100529_669x124_scrot

gtandersen:
Very nice, @kvviecien.

Issue #416, opened 2019-12-29
Add link to password reset on Runbox 7 log in page


davidbowdley:
Add link to password reset on Runbox 7 log in page. 

https://runbox.com/pw_reset

davidbowdley:
I think this trivial change should be implemented as soon as possible in order to help customers who are unable to log in due to a lost password.

gtandersen:
The link can be added to the error message displayed when the username/password combination is wrong.

gtandersen:
See also #525.

Issue #525, opened 2020-04-14
Migrate password reset form to Runbox 7


gtandersen:
The password reset form is currently only available in Runbox 6 (https://runbox.com/pw_reset) and should be available in Runbox 7 as well.

PR #596, opened 2020-05-20
feat(login): Add password reset link to login window

kvviecien:
When user gets an error while logging in, e.g. when putting in incorrect credentials, link to password reset page appears on the bottom.

Issue: #416
---
Also changed the position and styling of the error message and and the checkboxes. Looks simpler, more modern. Screenshots below:

runbox_old 
runbox_new


kvviecien:
I followed the suggestion of @gtandersen from Issue #416 about showing the link only when the username/password combination is wrong. But I thought maybe it would be better if the link was visible all the time? 

And I am so sorry, I just noticed @hernan604 was already assigned to it. Issue #525 looked very pleasing to me and I thought I would make some steps towards it, not looking at the assignees...

src/app/login/login.component.ts
```diff
@@ -135,7 +135,8 @@ export class LoginComponent implements OnInit {
         if (loginresonseobj.user_status > 0 && loginresonseobj.user_status < 5 && loginresonseobj.error) {
             this.accountExpired = true;
         } else {
-            this.login_error_html = '<p>' + loginresonseobj.message + ': ' + (loginresonseobj.error || 'error') + '</p>';
+            this.login_error_html = '<p>' + (loginresonseobj.error || 'Error occurred') + '</p>';
+            console.log(loginresonseobj);
```
tadzik, line 139: Leftover debugging code? :)

kvviecien: Thanks. Got rid of it :)

gtandersen:
Nice work @kvviecien (and I'm sure @hernan604 doesn't mind :).

I suggest only showing the link when the login fails as it takes up another line of vertical space, so the login box gets somewhat cramped depending on the window size.

kvviecien:
Thanks! It's good to go then :)

Issue #456, opened 2020-02-04
Sorting order change instead of opening an email


kotecky:
When the very bottom of a message list is clicked (the last visible message on page), the sorting order is change instead of opening the email. (FF under Win10)

PR #597, opened 2020-05-20
fix(canvastable): Make it possible to open email from the bottom

kvviecien:
The footer overlapped with message on the bottom, but wasn't visible.
Clicking on it resulted in reordering emails (just like you would click
on the corresponding header).
The footer has now 0 height, so it's not possible to click on it
anymore.

Fixes: #456


gtandersen:
Good catch!

PR #598, opened 2020-05-21
test(e2e): Skip the welcome dialog in compose tests

tadzik:
Also adds a delay to tests that need the mailviewer ready


tadzik:
#599 will hopefully do this, but better.

Issue #238, opened 2019-08-15
folder buttons should be links


kotecky:
When composing a message I often open the original message or a loosely related message in one window and composing the reply or new message in another window, side by side with the first one. Or I have two mail tabs open, for a number of other reasons. 

This is easily achieved by right-clicking a menu item (inbox, folder name, compose, drafts) and opening in a new window or tab. This would work for <a> elements, not for <span> elements used in Runbox7. 

I miss this possibility.

gtandersen:
I have a similar workflow, and an alternative would be to allow opening messages and Compose in new tabs/windows.

See also #429 and #614.

PR #599, opened 2020-05-21
feat(webmail): Give folders and messages URL fragments

tadzik:
This allows linking to specific folders/messages
as well as going back and forth between messages with browser buttons.

It also attempts to fix our e2e tests to make them pass (or fail) more consistently.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -30,7 +30,7 @@ import { MatIconRegistry } from '@angular/material/icon';
 import { MatSidenav } from '@angular/material/sidenav';
 import { MatSnackBar } from '@angular/material/snack-bar';
 import { MoveMessageDialogComponent } from './actions/movemessage.action';
-import { Router, RouterOutlet } from '@angular/router';
+import { Router, RouterOutlet, ActivatedRoute } from '@angular/router';
 import { DomSanitizer } from '@angular/platform-browser';
 import { HttpClient } from '@angular/common/http';
 import { MessageTableRow, MessageTableRowTool } from './messagetable/messagetablerow';
@@ -138,6 +138,7 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
     public snackBar: MatSnackBar,
     public dialog: MatDialog,
     private router: Router,
+    private route: ActivatedRoute,
     public progressService: ProgressService,
     private mdIconRegistry: MatIconRegistry,
     private http: HttpClient,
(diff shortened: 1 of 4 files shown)


tadzik:
https://github.com/runbox/runbox7/issues/238 should be a consideration here: right now all the links are created on-the-fly on the JS side, making it impossible to open things in new tabs.

gtandersen:
Nice @tadzik!

petersalomonsen:
the cypress tests look very efficient, and I like the introduction of URL fragments for folders and messages.

src/app/app.component.ts
```diff
@@ -151,24 +152,27 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
...
-              }
-              if (combo === 'j') {
-                  this.canvastable.scrollDown();
-              }
-              const e: ExtendedKeyboardEvent = event;
-              e.returnValue = false;
-              return e;
-          }));
+    private _hotkeysService: HotkeysService
+  ) {
```
castaway, line 156: Is this change of indentation on purpose?

tadzik: It is – I switched it to 2-spaces to be like the rest of the file, since it was off before that.

castaway: Its still 4 spaces...

tadzik: Well, it's 4 spaces from column 0, but 2 from the outer scope... I think? :)

The empty line is silly indeed, I can remove it.

src/app/app.component.ts
```diff
@@ -577,19 +598,20 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
       this.selectedRowIds[this.selectedRowId] = true;
 
       if (!multiSelect && columnIndex > 1) {
+        let messageId: number;
         if (this.showingSearchResults) {
-          this.singlemailviewer.expectedMessageSize = this.searchService.api.getNumericValue(this.selectedRowId, 3);
```
castaway, line 581: Confused here where expectedMessageSize went ?

tadzik:
Replaced by https://github.com/runbox/runbox7/pull/612

src/app/app.component.ts
tadzik: Hopefully described a bit better in https://github.com/runbox/runbox7/pull/612/commits/75616816347eb35ad40e5dc71479ba1053036c72 :)

src/app/app.component.ts
tadzik: Commited more explicitely in https://github.com/runbox/runbox7/pull/612/commits/ad923afeb1f09495749d1ad37b164044c3b04900

Issue #441, opened 2020-01-24
Message list doesn't change format when column changes size on mobile devices


davidbowdley:
On an iPad (haven't tried other mobile devices) the message list does not change format when its  column is made narrower.
![message_list_column](https://user-images.githubusercontent.com/5822260/73076693-83893a00-3eb6-11ea-92e1-c2de3b52ef09.PNG)

gtandersen:
@davidbowdley From the screenshot it looks like the message list is already in a wrapped format. More of it would be visible if the folder pane was collapsed, which should happen automatically if the screen size is small (enough).

Can you describe in more detail what you would like to see here?

davidbowdley:
@gtandersen Yes, it seems you are right. I will need to go back and check what was different here again. I may have described the discrepancy in the wrong way.

kvviecien:
![runbox ipad pro](https://user-images.githubusercontent.com/24294620/82600057-8fa7ae80-9bad-11ea-97b4-167eec630a1f.png)
![runbox ipad](https://user-images.githubusercontent.com/24294620/82600059-90404500-9bad-11ea-8008-9391cfb05281.png)

On iPads with smaller screens (768px) everything behaves like it should, but on iPad Pro for example (1024px) the site gets styling for desktops and everything gets squished together. I think the mobile threshold should be moved up just a little, to also include a tad bigger screens.

Working on it!

davidbowdley:
Talking to a number of people I get the impression there is a range of screen sizes (maybe only a narrow range) where the desktop view is presented and then comments are made about the message pane being very narrow at the bottom of the screen. So yes, it could be that the threshold for mobile view needs to be changed. It needs more testing though.

PR #600, opened 2020-05-21
fix(styling): Fix breakpoints for iPad Pro

kvviecien:
Now it behaves like with other mobile devices, which isn't resulting in squishing all the content together side by side.

Fixes: #441

runbox ipad pro fix


gtandersen:
Nice @kvviecien -- sometimes 1px is all it takes. :)

Issue #542, opened 2020-04-20
Close email when unchecking


vertigo220:
When unchecking/deselecting an email, i.e. clearing the check box that's automatically checked when an email is opened, it should optionally close the email, as an alternate way of doing so vs clicking the x button at the top-right.

gtandersen:
@vertigo220 How would you suggest that option be presented?

vertigo220:
Are you asking how it should be done or how it should be offered to the user as an option to enable or disable, i.e. choose whether they want that behavior?

tadzik:
This brings up a good point, and a somewhat annoying discrepancy between what is selected and what is displayed.

When an email opened (by clicking on it), it also becomes selected. However, selecting an email does not open it.

When an email is closed, it becomes de-selected. But de-selecting an email, as @vertigo220 says, it does not become closed.

With multiple selections it gets outright awkward: while an email is opened you can select more emails: but only the original one is displayed then. When you close it, everything becomes de-selected -- so maybe it should not have been selectable in the first place?

I suggest we agree on some consistency here and hopefully eliminate the unexpectedness that @vertigo220 experienced -- while technically we don't have a feature that would close the email when it's de-selected, the UI seems to heavily imply that this is how it ought to work.

tadzik:
A possible solution (design-wise, aside of implementation details) would be to always have an email being selected mean that the same email is being displayed – and vice versa. This cuts all the confusion when a single email is involved, but may become problematic with multiple emails. We could display a summary of the selected ones on the email preview pane, possibly with some message actions: that would be even handy. But what if the user is not using the always-open preview pane and also uses the horizontal preview? Then as soon as they start selecting multiple emails to move/delete, a mailviewer opens up, takes up half the screen space and maybe even interrupts the selection process itself (at the very least it becomes confusing and annoying). I'm not sure if solving this elegantly is possible without a careful redesign of the webmail layout and its dynamicness – but maybe it's worth it. We could also distinguish between selecting a single email and selecting multiple (so: only opening the mailviewer when the mouse button is released). This avoids the issue of the horizontal mailviewer popping in mid-selection, but still has a flaw of taking away the screen estate that the user might need to keep selecting more stuff.

Ideas?

castaway:
My inclination would be "dont check the selected box when an email is opened".. keep that for operations across multiple emails..

castaway:
[...]

Issue #303, opened 2019-09-19
checkboxes toggled in folder with message open are cleared upon message close


deutrino:
STEPS TO REPRODUCE

1. Open a folder.
1. Open a message.
1. Toggle some checkboxes in the folder with the message open.
1. Close the message.

EXPECTED RESULT

The checkboxes' toggle state is not changed.

ACTUAL RESULT

The toggled checkboxes are cleared.

gtandersen:
@deutrino This seems to be a duplicate of #10.

tadzik:
Somewhat related to https://github.com/runbox/runbox7/issues/542.

vertigo220:
Agree this is a duplicate of #10.

gtandersen:
This is no longer reproducible; closing issue.

Issue #11, opened 2018-12-15
Show message action buttons when selecting messages with preview window open


gtandersen:
When a message is opened and shown in the preview window, the message action buttons aren't shown when messages are selected in the list.

Ref: https://community.runbox.com/t/selecting-scrolling-issues/670

kotecky:
BTW, when rendered on a large screen, the buttons could be located at the edge of the message list and not far right by at the edge of the message preview pane. Which is particularly painful on a laptop with a touchpad.

vertigo220:
Agree 100% with both of these. On a large/wide screen, having the buttons all the way on the right is inefficient, requiring the user to move back and forth quite a bit. The buttons definitely need to show when the preview pane is showing, and they should ideally be in a better location, which would also help with the first issue. They should probably be either at the top or bottom of the message list, or maybe even between the folder and message list.

gtandersen:
This is fixed.

Issue #10, opened 2018-12-15
Keep message selections when closing preview pane


gtandersen:
"Closing message pane deselects multiple selection in message list - selection should persist."

Ref: https://forum.runbox.com/newreply.php?tid=56, https://community.runbox.com/t/selecting-scrolling-issues/670.

kotecky:
And similarly: previewing a message should not deselect the selection either. (Typically used to verify whether a given message should or should not be part of a selection that is being made.)

vertigo220:
It's disappointing to see this issue, which should be _relatively_ simple to fix and is, IMO, quite serious (imagine going through and selecting a bunch of emails, then closing the preview pane and losing all your selections; I would be extremely upset), has been around for almost 1.5 years and hasn't been fixed yet.

Furthermore, this issue is even more serious due to issue #11, since the user is forced to close the pane in order to gain access to those functions.

vertigo220:
There has been some discussion on this in #542, which I've closed since it's along the same lines as this and it's best to just reference it and carry over the important parts into this issue. That issue started by seeing this behavior from a different angle, essentially in reverse. Instead of (well, in addition to really) finding it problematic that message selections are removed when the preview pane closes, I thought it would be nice if unchecking an email would close it. However, as @tadzik pointed out, the current design is contradictory and leads to unexpected behavior, and it's probably best to just disconnect the two functions (checking/unchecking and opening/closing messages) so one has nothing to do with the other.

There's also some discussion about how to handle multiple message selections:

"with multiple emails. We could display a summary of the selected ones on the email preview pane, possibly with some message actions: that would be even handy."

and

"each checked email could show as a narrow strip with just the header info (from/to/subject/date) in the preview pane, making it easy to see what emails are selected as you scroll through the folder. Sort of like how Fastmail does it but not stacked."

@tadzik brought up concerns about this showing and causing disruption if the user doesn't normally use the preview pane, to which I would say there could simply be a separate option to enable this, which would automatically be enabled if the preview pane is in use and automatically disabled with the ability to enable it if not.

castaway:
Re: summaries of whats selected, does turning on the "Inline previews" help?

vertigo220:
[...]

PR #613, opened 2020-05-27
fix(webmail): Display selected-mail operations whenever more than one…

castaway:
… message is selected

Fixes several reported issues, where selected row checkboxes will
uncheck when closing an open message. Also selecting multiple messages
with the message pane open, would not display the operations buttons
for acting on the selected emails.

Fixes (parts of) #10,#11,#303

* [x] Currently, opening an email (clicking on subject for example, not the checkbox), also "selects" the related checkbox. In order for this to not look strange, I made the "select operations" only appear when more than one email was selected - FIXED (see #542)
* [x] This fails existing tests, as we have one that  checks the "move" button appears when only one email is selected - FIXED (by the above)
* The changes made make the "select operations" buttons appear regardless of whether the mail view pane is open, closed, or being closed after it was previously open, as requested in the tagged issues.
* NB: These changes undo some obviously deliberate previous choices by @petersalomonsen (back before the code was on github), so we should be sure how we want the UI to act.
* [x] Possibly #542 should be rolled into this, and we should de-couple "view email" with "select emails for multi operations" - DONE
* This is because this PR (currently) makes things a bit weird if you:
    * View an email (thus opening the view pane, and causing the checkbox to be checked)
    * check some more checkboxes, up to / including unchecking the viewed email
* [x] We're left in a state where its not obvious which email is opened, as highlighting is applied to the checked rows, but nothing distinguishes the viewed email .. Extra highlight added

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -233,7 +233,7 @@
   <router-outlet (activate)="hasChildRouterOutlet=true" (deactivate)="hasChildRouterOutlet=false"></router-outlet>
 </mat-sidenav-container>
 
-<div class="contextToolButtons" *ngIf="!hasChildRouterOutlet && showSelectOperations && !singlemailviewer.messageId">
+<div class="contextToolButtons" *ngIf="!hasChildRouterOutlet && showSelectOperations">
   <button mat-fab matTooltip="Move&nbsp;to&nbsp;folder" (click)="moveToFolder()">
     <mat-icon>folder</mat-icon>
   </button>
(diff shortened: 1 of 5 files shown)


castaway:
Any thoughts on how to make it more obvious which email is open (now that its not the checkbox).. row highlight?

gtandersen:
@castaway Yes; that is already the case though?

castaway:
> @castaway Yes; that is already the case though?

we currently highlight the currently active (clicked on etc) email.. when having one open, and selecting checkboxes, it becomes not the same thing as "the open one". Probably easiest if we install this PR somewhere you/Dave can poke at it?

gtandersen:
Looks good @castaway! Can you have a look as well, @tadzik?

Issue #455, opened 2020-02-04
Date/times column in message list after midnight


kotecky:
Funny oddity rather than a bug ;) 

Date column in a message list displays a date for older messages and the time for those arrived today. Once midnight passes (if app page not refreshed?), the time is kept for the messages for the day that once was "today" and the ones arrived after midnight are displayed with a date (ipo of time). 

![dates](https://user-images.githubusercontent.com/2297444/73734993-d7114880-473e-11ea-9782-cf47aefa659d.png)

gtandersen:
@kotecky Thanks for discovering and describing this issue -- I have also noticed it on occasion but not associated it with the time of day/night.

gtandersen:
Reopening this as it occurring still/again:

![Uploading Screenshot 2023-09-12 at 08.52.22.png…]()

PR #615, opened 2020-05-28
fix(messagetable): Display time instead of the date for messages received after midnight

kvviecien:
If the last time user refreshed the page was before midnight, messages they receive after midnight will now have the time shown in the table instead of today's date (that is different than the date when user refreshed the page).

Fixes: #455


tadzik:
We may want to consider changing the previous day messages back to the date format: otherwise after a few days of runtime the only thing shown will be the hours :)

Issue #508, opened 2020-04-01
When closing a draft the text does not stop where you can see the window ends and extends way beyond so that all the message is visible.


davidbowdley:
When closing a draft the text does not stop where you can see the window ends and extends way beyond so that all the message is visible.

I can't supply the message I saw this with as it is confidential, but I have sent it to @gtandersen 

It's clear where the message preview should end as you can see the normal border that would contain the text, but the text just continues on. Refreshing the page resolved the issue until you open and close the draft again.

PR #618, opened 2020-05-29
fix(draftdesk): Fix text overflowing after closing draft

kvviecien:
Overflow is hidden now, only 2 lines of the email are shown in the preview.
Removed blank spaces between the drafts and improved styling in general. Drafts now take up the entire width and the look throughout the page is more consistent.

Fixes: #508

### DIFF
--- a/src/app/compose/compose.component.scss
+++ b/src/app/compose/compose.component.scss
@@ -3,3 +3,14 @@
     border: 1px solid #949494;
     padding: 10px;
 }
+.draftPreview {
+    max-height: 100px;
+    overflow: hidden;
+
+    span {
+        display: -webkit-box;
+        -webkit-line-clamp: 2;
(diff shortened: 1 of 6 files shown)


src/app/compose/compose.component.scss
```diff
@@ -3,3 +3,75 @@
...
+    align-items: center;
+    justify-content: flex-end;
+    margin-right: 5px;
+}
+
+.draftPreview {
+    max-height: 100px;
+    overflow: hidden;
+
+    span {
```
tadzik, line 29: Huh, I didn't know CSS can do nested blocks. Or is it just scss?

kvviecien: That's SCSS, pure CSS can't do that unfortunately :(

Issue #603, opened 2020-05-25
Contacts: deleted contact cards stay visible/searchable until refresh


kotecky:
When a contact card is deleted, it remains in the list of contacts and can be found by searching and can be viewed. 

Still cached? Ideally, a deleted card should be deleted from the local data too.

castaway:
Change is now live, please retest?

kotecky:
Works fine now. Thanks!

Issue #595, opened 2020-05-20
Cache contacts locally for use with Compose


gtandersen:

tadzik:
I couldn't reproduce the slowness you mentioned, but noticed something relevant (and possibly related):

Opening Runbox7 in a new browser, compose doesn't autocomplete anything at all. Even after going to contacts (and verifying that all contacts are indeed there) and back, nothing shows up in compose at all. Only after enabling the search index contacts get autocompleted ­– the ones from the Contacts, that is. So it seems like with the absence of searchindex, the compose form doesn't search the searchindex (obviously), but then it doesn't even attempt to gather results from contacts.

Could this be the same issue you noticed, or is there more to investigate here?

gtandersen:
@tadzik I realized from your comment that the recipient list is (still) built from the index and not the Contacts database, and that probably explains the issue as I have upwards of 400k messages in the index. 

The solution seems to be switching to the Contacts list if feasible. :)

kotecky:
> The solution seems to be switching to the Contacts list if feasible. :)

A user's comment on this: Using the Contacts is of course a necessity, using an index based on the in/out-box is a good feature. I guess the users will prefer a mix of both, and ideally displayed not in an alphabetical order but taking into account the frequency of use: when typing in letters of a name. Often when typing a contact's name, other names much less used are always sitting on top. This is annoying mainly on a mobile phone where there is less space in the interface and one needs to scroll. 

One danger of the index of in/out-box is the presence of addresses that do not exist (closed accounts, typos) and irrelevant ones.

castaway:
> @tadzik I realized from your comment that the recipient list is (still) built from the index and not the Contacts database, and that probably explains the issue as I have upwards of 400k messages in the index.
> 
> The solution seems to be switching to the Contacts list if feasible. :)

It is actually built from both, but only updates from either if the search index changes - I'm working on separating that, so that its updated if/when either changes.

gtandersen:
@castaway What is the status of this issue? I think it would be useful if the CardDAV database is queried and displayed first, instead of waiting for the search index since that can take longer.

castaway:
The separation is complete.. the list is now composed of both contacts, and index recipients, and updated when either one changes (not just when/if the index changes)

gtandersen:
@castaway This doesn't seem to improve the responsiveness of the recipient autocomplete though. There is still considerable latency with a sizeable index.

castaway:
[...]

Issue #552, opened 2020-04-20
Contact list used in compose isn't refreshed when new contacts are added


tadzik:
Compose autocomplete only contains contacts fetched at app startup, not ones added later.

Steps to reproduce:

1. Go to Contacts, add a new contact
2. Go back to webmail, compose new email
3. Newly added contact doesn't show up in recipients autocomplete
4. Refresh the page
5. Contact now autocompletes as it should

tadzik:
Possibly related to https://github.com/runbox/runbox7/issues/595: perhaps contacts are only loaded when the search index is *first* loaded?

PR #623, opened 2020-06-04
Update contacts and search index recipients caches separately, remove duplicate ContactsService instance

castaway:
Sorting out several issues with updated/deleted contacts not being reflected in the contacts list (side menu), or in the webmail "compose" window until some time after the update has been made.

Fixing #552, #595, #603


src/app/compose/recipients.service.ts
```diff
@@ -48,48 +50,70 @@ export class RecipientsService {
...
+                () => this.recipients.next([])
+            );
+        });
+    }
+
+    private updateRecipients(groups: Recipient[] = []) {
+        const uniqueRecipients: {[uniqueKey: string]: Recipient} = {};
+        this.recipientsUpdating.forEach(entry => {
+            if (!uniqueRecipients[entry['uniqueKey']] ||
+                entry['origin'] === 'searchindex') {
```
tadzik, line 112: These are set to either `'contacts'` or `'searchindex'` in other parts of the code – so this part of the `if` will always eval to false, I think.

Perhaps it'd be better to switch to an enum for these, to avoid typos like this :)

It also seems like in its current form (if corrected), it'd make searchindex win over contacts when updating the list – is that intentional, and/or covered by our tests?

castaway: Oops, yup that was me being a nitwit (tests pass tho!) .. enum sounds like a good idea!

PR #628, opened 2020-06-05
Hernan604/rmm7/list aliases where user is owner

hernan604:


src/app/rmmapi/rmmhttpinterceptor.service.ts
```diff
@@ -70,6 +72,8 @@ export class RMMHttpInterceptorService implements HttpInterceptor {
                             console.log('proceed with 2fa login');
                         } else if (req.url === '/rest/v1/me') {
                             this.authguardservice.redirectToLogin();
+                        } else if ( req.url.match(/^\/rest\/v1*/g) && r.body.errors ) {
+                            this.show_error(r.body.errors.join(''), 'Dismiss');
```
tadzik, line 76: Please don't do this in httpinterceptor. There's already places in the code that react to api errors, and this will display an *additional* snackbar.

hernan604: The problem is, this intercepts the error and the useragent that made the request does not receive the error because it ends up in the else. So im unable to receive the error

hernan604: i mean, without this piece of code, it seems i cant get the error ?

tadzik: The interceptor doesn't consume the error, just like it doesn't consume the non-errors: it merely looks at them and passes them forward unmodified. Otherwise stuff like https://github.com/runbox/runbox7/blob/master/src/app/rmmapi/rbwebmail.ts#L293 won't work.

If your error gets lost somewhere then it must be something else's fault.

hernan604: i removed it now and was able to display the error
im not sure if i couldnt get the error, or, i added the error handling in the interceptor so it would handle the display of errors in a single place
but lets remove it then
if ( reply.data.status === 'error' && reply.data.errors.length ) { this.show_errors( reply.data.errors.join(" "), 'Dismiss' ) }
and then we never need to handle that else where
maybe that was the idea

Issue #505, opened 2020-03-28
Emtpy error message when trying to create a foldere that already exists


davidbowdley:
I came across an error where if you try to create a folder with a name of a folder that already exists a notification pops up at the bottom of the screen that is empty. I suspect it is trying to say that the specified folder already exists.

tadzik:
I couldn't reproduce it: but the error message is awful nonetheless :D

![2020-03-31-154412_349x49_scrot](https://user-images.githubusercontent.com/86378/78033334-8b59d480-7366-11ea-893c-1c676b3da56a.png)

Can you give me a Steps to Reproduce that makes it show an empty error? Is there anything in the JS console when that happens?

davidbowdley:
With Inbox highlighted as the active folder I tried to create a folder called Spam as well as a custom folder that already exists and I got the blank error. I wonder if it is browser specific. Let me check.

davidbowdley:
It doesn't appear to be browser specific. I've now got the same result in Firefox and Safari. I can't see anything particularly relevant in the web console or browser consoles, but I am not sure what I would be looking for.

kvviecien:
The weird thing is that on production (runbox.com) the error message is empty, but when running the server from master locally, I get "folder_exists" in the message. Trying to figure it out.

kvviecien:
Now the formatting is better, more readable to the user, but I really don't know why an error message appears when run only locally (for me at least). All the logs and results from API look the same so I really can't tell what's going on.

PR #629, opened 2020-06-05
fix(errors): Better formatting of backend errors

kvviecien:
Instead of showing an error message like "folder_exists", it will show a formatted message "Folder exists"

Issue: #505

### DIFF
--- a/src/app/rmmapi/rbwebmail.ts
+++ b/src/app/rmmapi/rbwebmail.ts
@@ -295,9 +295,10 @@ export class RunboxWebmailAPI {
             if (res.status === 'error') {
                 if (res.errors && res.errors.length) {
                     const error_msg = res.errors.map((key) => {
-                        return this.rblocale.translate(key);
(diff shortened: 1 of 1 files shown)


tadzik:
> I was wondering why the checks aren't passing but it says that "Commit message for [f6adf35](https://github.com/runbox/runbox7/commit/f6adf350f0c9f472ffd5c28bd0644c93984dd276) doesn't have the correct format" which is @tadzik [commit](https://github.com/runbox/runbox7/commit/f6adf350f0c9f472ffd5c28bd0644c93984dd276). Why is it an issue here?

That'd be Travis being dumb. The [test](https://github.com/runbox/runbox7/blob/master/policy-tests/check-commit-log.js#L3) is told to skip merges, so commits like these should never be inspected in there. It all passes locally, so I'll just go ahead and merge it.

tadzik:
(For future reference: the problem remained on travis after the merge, but seems to be fixable by telling Travis to clear all its repository caches: it then re-clones from scratch and is confused no more.

PR #635, opened 2020-06-10
feat(compose): Show recently used recipients under input forms

tadzik:
TODO:

- [x] Tests
- [x] Fine-tune the amount of contacts suggested
- [x] Make layout more responsive (WIP in https://github.com/runbox/runbox7/pull/649)

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -35,6 +35,12 @@
       <mat-divider></mat-divider>
     </mat-nav-list>
 
+    <app-popular-recipients
+        (recipientClicked)="searchFor($event)"
+    ></app-popular-recipients>
+
+    <mat-divider></mat-divider>
+
     <rmm-folderlist #folderListComponent
         [folders]="displayedFolders"
         [folderMessageCounts]="messagelistservice.folderMessageCountSubject"
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -107,14 +107,14 @@
           *ngIf="showMultipleSearchFields && searchService.downloadProgress===null"
           [currentFolder]="selectedFolder"
           (close)="showMultipleSearchFields = false"
-          (searchexpression)="searchFieldKeyUp($event)">
+          (searchexpression)="searchFor($event)">
         </app-multiple-search-fields-input>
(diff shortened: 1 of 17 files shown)


tadzik:
One suggested feature was to allow blacklisting of some recipients – but I think that'll require some API support to make it sync between devices: otherwise it may end up more frustrating than useful.

src/app/app.component.html
```diff
@@ -35,6 +35,12 @@
       <mat-divider></mat-divider>
     </mat-nav-list>
 
+    <app-popular-recipients
+        (recipientClicked)="searchFor($event)"
```
castaway, line 39: Should this be "from:$event" ? Cos like this it finds mentions of the address inside other emails too

src/app/compose/compose.component.scss
```diff
@@ -75,3 +63,87 @@ button {
...
+    }
+}
+
+.draft-buttons {
+    display: flex;
+    align-items: center;
+    justify-content: flex-end;
+    flex-grow: 1;
+
+    #trashDraft { order: 1 }
```
castaway, line 120: is this part of a fix for a different issue?

castaway: .. nm Im looking at the whole thing at once

src/app/compose/compose.component.html
```diff
@@ -43,18 +39,55 @@
...
+        <div class="recipientSuggestionContainer" *ngIf="suggestedRecipients.length > 0 && editing">
+            <span>Recently used</span>
+            <div class="recipientSuggestionList">
+                <div *ngFor="let recipient of filteredSuggestions" class="suggestionItem">
+                    <mat-chip
+                        (click)="addRecipientFromSuggestions(recipient)"
+                        draggable="true"
+                        (dragstart)="recipientDragged($event, recipient)"
+                    >
+                        {{ recipient.nameAndAddress }}
```
castaway, line 71: I'm getting "null <address>" for those with no names :(

castaway:
On my initial load, I got "Popular recipients" listed on the main nav menu, but opening compose I got no "recently Used", they appeared on a page reload. Not sure if thats just me,

Looks great else!

Issue #236, opened 2019-08-15
comma in "from" results in two addresses when replying


kotecky:
If the name being part of an email address contains a comma: 
    ![image](https://user-images.githubusercontent.com/2297444/63087527-b7008980-bf52-11e9-8cb6-8d05c98b6048.png)

when replying to (or replying to all) the "to" input field contains two addresses, one of them being a non-sense: 
![image](https://user-images.githubusercontent.com/2297444/63087721-22e2f200-bf53-11e9-8c03-26bc52ed105e.png)

tadzik:
If I'm understanding [RFC2822](https://tools.ietf.org/html/rfc2822) correctly, commas are actually illegal in From headers unless wrapped in double quotes like so: `"User, Sicher 24" <info@bla bla>`. If that's how it was written in the original email, that means that we removed those somewhere during our handling of the message – but if that's how it arrived then this is actually a somewhat expected behaviour :­)

I'll test it on myself with some funky addresses and see how it turns out.

kotecky:
@tadzik would the original message header help?

tadzik:
@kotecky absolutely – if it contains anything you'd rather not publicize, feel free to email it to me at \<my github username\>@runbox.com

kotecky:
Emailed.

PR #638, opened 2020-06-12
#236 remove split email address lists on comma

castaway:
WIP:

Move lots of representations of email addresses / address lists (eg recipients) to be arrays of MailAddressInfo objects, so that code doesn't need to keep splitting / rejoining them using commas. Splitting on commas fails when the quoted name part contains a comma.

Changed "to" field so  far

TODO: "cc" and "bcc" fields


castaway:
This branch is built on top of @tadzik 's popular-recipients currently, so will need to wait on that finishing (or be rebased) before applying.

PR #643, opened 2020-06-18
236 remove split on comma

castaway:
All "TO" fields updated (and tests passing hopefully)

Also done: CC + BCC fields

### DIFF
--- a/src/app/compose/draftdesk.service.spec.ts
+++ b/src/app/compose/draftdesk.service.spec.ts
@@ -23,8 +23,8 @@ import { MailAddressInfo } from '../xapian/messageinfo';
 
 
 describe('DraftDesk', () => {
-    it('Reply: Address with object', (done) => {
-        console.log('Reply test: Address with object');
+    it('Reply: Address with object, single reply', (done) => {
+        console.log('Reply test: Address with object, single reply');
         const mailDate = new Date(2017, 6, 1);
 
         const timezoneOffset: number = mailDate.getTimezoneOffset();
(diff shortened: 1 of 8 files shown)


src/app/compose/draftdesk.service.ts
```diff
@@ -58,7 +58,7 @@ export class DraftFormModel {
         const ret = new DraftFormModel();
         ret.from = fromAddress.email;
         ret.mid = draftId;
-        ret.to = to;
+        ret.to = to ? MailAddressInfo.parse(to) : null;
```
tadzik, line 61: If we declare it as an array, and the default value is `[]` maybe we should default to `[]` over `null` here too?

castaway: Oops, yes seems saner

src/app/compose/draftdesk.service.spec.ts
```diff
@@ -33,56 +33,209 @@ describe('DraftDesk', () => {
...
-        draft = DraftFormModel.reply({
+        done();
+    });
+    it('Reply: Address with MAI', (done) => {
+        console.log('Reply test: Address with MAI');
+        const mailDate = new Date(2017, 6, 1);
+
+        const timezoneOffset: number = mailDate.getTimezoneOffset();
+
+        const timezoneOffsetString: string = 'GMT' + (timezoneOffset <= 0 ? '+' : '-') +
```
tadzik, line 152: This piece is duplicated in this file, could probably be refactored into a function.

I wonder, though, if we really need all this computing? Do the tests care for something more than "the timezone is the same before and after?"

tadzik:
Found an issue:

1) Open the CC form
2) Enter some email address
3) Close the CC form with the X
4) Open the CC form again
5) The previous email address is still there

That's because the X still resets the form contents, not the model itself.

src/app/compose/draftdesk.service.ts
```diff
@@ -112,7 +119,7 @@ export class DraftFormModel {
 
         if (!useHTML && mailObj.rawtext) {
             ret.msg_body = '\n' + mailDate.toISOString().substr(0, 'yyyy-MM-ddTHH:mm'.length).replace('T', ' ') + ' ' +
-                timezoneOffsetString + ' ' + sender + ':\n' +
+                timezoneOffsetString + ' ' + sender[0].nameAndAddress + ':\n' +
```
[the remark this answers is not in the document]
castaway: If sender is empty / not an array then several other things will break too - we dont currently check any of the incoming call arguments for sanity in this code..

src/app/compose/draftdesk.service.spec.ts
castaway: Good point, I just duplicated what was already there..

castaway:
> Found an issue:
> 
> 1. Open the CC form
> 2. Enter some email address
> 3. Close the CC form with the X
> 4. Open the CC form again
> 5. The previous email address is still there
> 
> That's because the X still resets the form contents, not the model itself.

Oops, now fixed - even less use of / reliance on the formControl objects (wonder if I can delete em yet)

PR #649, opened 2020-06-23
fix(compose): Make layout more responsive

kvviecien:
Better responsiveness and more organized layout for mobile

### DIFF
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -818,19 +818,6 @@ mat-expansion-panel {
     }
 }
 
-#buttonCC, #buttonBCC {
-    min-width: 50px !important;
-}
-
-.recipientSuggestionList {
-    display: flex;
-    align-items: center;
-}
-
-.recipientSuggestionList div {
-    padding: 5px;
-}
-
 /* Styles to decrease vertical spacing on Compose */
 
 compose .mat-card-actions {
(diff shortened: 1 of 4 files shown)


src/styles.scss
```diff
@@ -856,7 +843,7 @@ compose input.mat-chip-input {
 }
 
 compose .mat-standard-chip {
-    bottom: 6px !important;
+    // bottom: 6px !important;
```
tadzik, line 846: I think this commented-out bit can go now :)

kvviecien: Right, I was probably checking something out and forgot about it as always.

kvviecien:
In my opinion, the Send button being bigger looks silly right next to the other ones being different size. Like someone made a mistake styling it.
If it's necessary to make it stand out, I'd rather make it a different color. Maybe grey out the other ones, take off a little bit of their saturation, and leave the Send button like it is now.

kvviecien:
Reordered the buttons and moved HTML and Attach buttons closer to the textarea.

I noticed the attachments looked a little wonky on mobile like this \/ so I fixed them too!
Screenshot 2020-06-25 at 23 08 06

Result:
Screenshot 2020-06-25 at 23 04 04

kvviecien:
@gtandersen I thought putting Send as last would be better since it's best accessible when close to the edge of the screen, so maybe adding a little more space would be better if you think the buttons are too close.

gtandersen:
@kvviecien Alright, let's try that. If it doesn't help we can always make it a bit larger than the other buttons. ;)

PR #650, opened 2020-06-29
fix(reply):Reply to all uses reply-to header if present

### LINKED ISSUE #509: Reply to all does not include the "reply to" address in the recipients.
Using the reply to all feature ignores the "reply to" field sent with email messages and only includes the other recipients.

"Reply to all" should use the "Reply-To" header AND include all emails in the "To" and "CC" message headers.

castaway:
Fixes #509


tadzik:
So `reply-to` is an array, and we only use one the first address. While the header may only show up once, according to the RFC, so can To and Cc: and they can all contain multiple addresses.

I'm not sure what's coming from the API, but perhaps we should make sure that multiple reply-to addresses will also be handled correctly by this.

PR #652, opened 2020-06-29
feat(contacts): Implement a 3-column layout for contacts

tadzik:


castaway:
Couple of initial queries/confusions (I started with no groups) 
- "Show All Contacts" isnt obvious what its for, it should start greyed out if no groups / not currently viewing a group. 
- I created a group, it appeared left (not under the existing list in middle col) - bit confusion for people swapping from the existign version, maybe we should label the 2 sets / lists somehow?

tadzik:
Awkward scrolling is fixed, and this is ready for review (again :))

PR #653, opened 2020-06-30
fix(webmail): Only recount folder unread counts after content change

castaway:
Should help with 10sec pauses/slow down when folders with large
contents exist.


src/app/rmmapi/messagelist.service.ts
```diff
@@ -218,6 +218,8 @@ export class MessageListService {
         if (hasChanges) {
             this.messagesInViewSubject.next(this.folderMessageLists[this.currentFolder]);
             this.refreshFolderList();
+            console.log('New/Moved messages, updating folder counts');
```
tadzik, line 221: I assume this is left out intentionally, to ease the debugging in near future?

castaway: Yes that was my thought

PR #671, opened 2020-07-16
(Gr)avatars

tadzik:
Implementation of contact pictures, sourced from either CardDAV contacts or gravatar.com.

Comments on designs and others welcome.

## Design

1. When viewing emails, all available avatars (except of the user's) are shown in the email header:

avatarbar-collapsed

1b. Hovering over the "collapsed" avatars fluently "unrolls" them and shows them side-by-side:

avatarbar-expanded

2. Contact details show the contacts' vCard picture if available:

2020-07-17-163914_508x183_scrot

2b. vCard avatar can be replaced or removed when clicked:

2020-07-17-164005_222x255_scrot

2c. Remote avatars can be replaced, but cannot be removed; instead they indicate where they're from:

2020-07-17-164105_254x253_scrot

## Considerations

Remote avatars are fetched dynamically from gravatar.com (currently). Knowing some of our users, this could be seen as a privacy violation, since an external service is contacted, and potentially leaks the fact that a particular email was opened (if a unique address is used). We could proxy gravatar.com queries through our own servers (which would also enable us to cache them more aggresively, over gravatar's 5 minutes `Cache-Control`), but that still doesn't eliminate the second issue. Whether that's serious enough to disable this feature by default is up for consideration. @gtandersen? @davidbowdley?

## TODO:

- [x] Show vCard avatars in mailviewer if available
- [x] Un-stub the contact details picture editing
- [x] Add a setting to disable the feature partially/entirely


gtandersen:
@tadzik This looks excellent. 

To the privacy considerations, the solution could include options to:

1. Disable the feature entirely.
2. Enable the feature using local (Runbox) data only.
3. Enable the feature using local and external data.

Option 2 could then be set as the default, and the menu could be expanded with a section to change between all three.

castaway:
I like the look of this, couple of thoughts re usability:

* if you start on the Contacts page and see an "upload picture", there's no indication anywhere that gravatars are also available - maybe a tooltip on the upload picture could indicate that?
* Maybe add the settings gear to the top of the contacts left nav menu too, so that users dont have to hunt for it
* what does/should "disabled" do on the contacts pages?

tadzik:
Exposed avatar settings in Contacts settings, disable now disables them in contact details as well.

tadzik:
Added a tooltip in 1366014:

2020-07-31-120859_264x180_scrot

There's now a lot of duplicated code related to appsettings I'd like to refactor before the merge :)

tadzik:
TODO: make it so that changing avatar settings immediately alters the state of mailviewer (without the need to switch to a different email and back to see the changes)

tadzik:
If someone wants to (re)review 9cad883..eb2ed6b, go ahead :) I'll merge&deploy on monday if nothing comes up.

PR #677, opened 2020-07-21
Webmail settings, disabling recipients list

tadzik:
Looks like this (when opened):

2020-07-21-132559_327x128_scrot

We could eventually add more things to it (maybe index synchronization as well?), and also expose the same thing as a full-page settings editor, with detailed explanations and such.

### DIFF
--- a/src/app/app-settings.ts
+++ b/src/app/app-settings.ts
@@ -1,0 +1,28 @@
+// --------- BEGIN RUNBOX LICENSE ---------
+// Copyright (C) 2016-2020 Runbox Solutions AS (runbox.com).
+//
+// This file is part of Runbox 7.
+//
+// Runbox 7 is free software: You can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the
+// Free Software Foundation, either version 3 of the License, or (at your
+// option) any later version.
+//
+// Runbox 7 is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
+// ---------- END RUNBOX LICENSE ----------
+
+export interface AppSettings {
+    showPopularRecipients: boolean;
+}
+
+export function defaultAppSettings(): AppSettings {
+    return {
+        showPopularRecipients: true,
+    };
+}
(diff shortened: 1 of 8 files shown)


castaway:
Could do with a tooltip on the gear icon.. maybe "Manage Display" or summat?

src/app/app.component.html
```diff
@@ -36,6 +55,7 @@
     </mat-nav-list>
 
     <app-popular-recipients
+        [ngStyle]="{ 'display': webmailSettings.showPopularRecipients ? 'block' : 'none' }"
```
castaway, line 58: Why a style and not an ngIf? (just curious)

src/app/compose/compose.component.html
```diff
@@ -62,7 +62,10 @@
                 id="buttonBCC"
             > BCC </button>
         </div>
-        <div class="recipientSuggestionContainer" *ngIf="suggestedRecipients.length > 0 && editing">
+        <div class="recipientSuggestionContainer"
+           [ngStyle]="{ 'display': webmailSettings.showPopularRecipients ? 'block' : 'none' }"
```
castaway, line 66: same question re ngIf

src/app/app.component.html
tadzik: ngIf is actually noticably slower when re-enabling it (takes a few hundred ms, seemingly), so I'd say the CSS is probably worth keeping for the sake of apparent "snapiness" of things. @gtandersen, what do you think?

gtandersen:
Nicely done @tadzik! 

Agree that moving the index synchronization control to this menu could be a good choice at this point, and that any delay when enabling/disabling options should be avoided if possible.

src/app/app-settings.ts
```diff
@@ -0,0 +1,28 @@
...
+// Runbox 7 is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
+// ---------- END RUNBOX LICENSE ----------
+
+export interface AppSettings {
```
gtandersen, line 20: Are we sure we want to call it AppSettings when it's strictly settings for the Webmail only (realizing it's a consequence of Webmail being the default component of the app)?

tadzik: I was forward-thinking to the times when this could also be referring to things not necessarily webmail-specific – gravatars, for instance, could also go here, and impact both the webmail and the contacts. I know forward-thinking is not very YAGNI, but this is probably coming soon (as soon as one of these ends up on master :))

gtandersen: @tadzik Good point. We do intend to integrate further non-webmail functionality so this takes that into consideration.

PR #684, opened 2020-07-23
fix(webmail): Fix switching folders not working in some cases

tadzik:
This includes things like going Inbox->Drafts->Inbox, /welcome->Inbox
and probably a few others. Also makes the selected section actually
hilight correctly, even if it's not a proper folder.

### DIFF
--- a/e2e/cypress/integration/folder-switching.ts
+++ b/e2e/cypress/integration/folder-switching.ts
@@ -1,0 +1,36 @@
+/// <reference types="cypress" />
+
+describe('Switching between folders (and not-folders)', () => {
+
+    function goToInbox() {
+        cy.get('rmm-folderlist mat-tree-node:contains(Inbox)').click();
+        cy.url().should('be', '/#Inbox');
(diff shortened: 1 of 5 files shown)


castaway:
Not quite: Click on Compose, and then on a Popular Recipient: -> 
```
sentry-error-handler.ts:30 TypeError: Cannot read property 'indexOf' of null
    at SearchService.push../src/app/xapian/searchservice.ts.SearchService.getCanvasTableColumns (searchservice.ts:1229)
```

tadzik:
> Not quite: Click on Compose, and then on a Popular Recipient

Good catch. I fixed that in 10e27fb, though I'm not super happy about the ~~hack~~ code – I'm open for alternatives.

No tests, as the mockserver doesn't support saved indexes yet, and thus popular recipients are not usable in e2e tests :( It'll probably be easier to implement custom saved searches (which will use the same codepath).

tadzik:
Bump? :)

PR #692, opened 2020-07-28
Activity indicators

tadzik:
This implements activity indicators, previously exclusive to Calendar as a generic component usable with anything within the app, with Contacts as the case study. It also introduces "multipart" activities that calendar didn't need, but contacts uses for things like deleting multiple contacts:

indicators

### DIFF
--- a/src/app/calendar-app/calendar-app.component.html
+++ b/src/app/calendar-app/calendar-app.component.html
@@ -178,8 +178,10 @@
 
             <mat-divider></mat-divider>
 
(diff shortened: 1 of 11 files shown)


tadzik:
Needs a sensible solution to indicate deleting multiple contacts – possibly as an ongoing activity, indicating "Deleting contacts (3/7)" or something along those lines. Will probably require complicating the API a little.

tadzik:
> Note that @hernan604 also started reusable components in the `account_security_v2` branch in `src/app/runbox-components`.

Good to know – we can merge them after both branches are merged :)

PR #699, opened 2020-08-04
Hernan604/account security v2 final2

hernan604:


src/styles.scss
```diff
@@ -139,11 +157,12 @@ mat-grid-tile.tableTitle {
 .contentButton {
```
tadzik, line 157: This would change these buttons globally, across the entire app. Perhaps it'd be better to create a new CSS class for this if you want a white button with bigger font?

src/app/runbox-components/runbox-slide-toggle.ts
```diff
@@ -0,0 +1,85 @@
...
+  ViewChild,
+  AfterViewInit,
+  ContentChild,
+  ElementRef,
+  TemplateRef,
+} from '@angular/core';
+import { CommonModule } from '@angular/common';
+import { FormsModule } from '@angular/forms';
+
+import { MatCardModule } from '@angular/material/card';
```
tadzik, line 35: Importing Modules doesn't do anything ­– you can safely throw all of these out :)

src/app/runbox-components/runbox-slide-toggle.ts
```diff
@@ -0,0 +1,85 @@
...
+        <ng-content></ng-content>
+        </mat-slide-toggle>
+    </div>
+    `
+})
+
+export class RunboxSlideToggleComponent {
+  @Input() is_checked = true;
+  private dialog_ref: any;
+  constructor(public dialog: MatDialog,
```
tadzik, line 77: I don't think this needs Dialogs or RMM?

backend-proxy-remote.conf.js
```diff
@@ -1,5 +1,5 @@
 let backendhostname = process.env.RUNBOX7_ANGULAR_BACKEND_HOST;
-let backendhostsecure = true;
+let backendhostsecure = false;
```
tadzik, line 2: Commited accidentally?: )

src/app/app.module.ts
```diff
@@ -121,6 +125,7 @@ const routes: Routes = [
   imports: [BrowserModule, FormsModule,
     HttpClientModule,
     HttpClientJsonpModule,
+    MatTabsModule,
```
tadzik, line 128: Those two new modules aren't used anywhere in the main app, so here they'll just make the compilation longer (or the bundle larger).

src/app/runbox-components/runbox-slide-toggle.ts
hernan604: fixed

tadzik:
Was replaced by https://github.com/runbox/runbox7/pull/735

PR #703, opened 2020-08-05
feat(app): Speed up the way sidemenu appears on the webmail page

tadzik:
This brings its behaviour in sync with contacts and calendar – sidepane being visible instantly (if in desktop), and still responsive.

The price we way is the animation of the sidemenu sliding in – which may be seen as nice or unnecessary, depending on your point of view (and the performance of your computer). To be discussed, @gtandersen @davidbowdley.


tadzik:
@gtandersen having tested that (iirc), what do you think?

gtandersen (approved):
Great @tadzik!

Issue #701, opened 2020-08-04
 /app/account page is empty


castaway:
This is what happens when I poke around in the urls - should it redirect to /account/components?

gtandersen:
Yes, it should redirect to `/app/account/components`.

tadzik:
I made it behave the same as the /components page (without the redirect) in the PR above ^

PR #724, opened 2020-08-14
fix(account): Make /account show the same thing as /account/components

tadzik:
Maybe one day it will be some generic account hub, but right now let's at least make it do something useful.

Fixes GH-701


castaway (approved):
Looks good to me

Issue #705, opened 2020-08-05
Stop first character in username on log in page from being uppercase on mobile devices


davidbowdley:
On mobile devices stop the first character in username on the log in page from being uppercase.

Issue #706, opened 2020-08-05
Use/set appropriate keyboard on log in page on mobile devices


davidbowdley:
On mobile devices use appropriate keyboards for username and 2FA entry.

For example, there is an option to show a keyboard with the @ symbol for usernames and for things like TOTP a numeric only keypad can be shown.

PR #726, opened 2020-08-15
feat(login): Set inputmode to show email and numeric keyboards on mobile as appropriate

gcjensen:
Closes https://github.com/runbox/runbox7/issues/705 & https://github.com/runbox/runbox7/issues/706.

I've opted to set just `inputmode` on the number inputs, rather than `type="number"` as well, as this has accessibility problems (among others). See https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/.

Although I followed [CONTRIBUTING.md](https://github.com/runbox/runbox7/blob/master/CONTRIBUTING.md), I've not updated the changelog as it appears this hasn't been done in some time, so doing so would have included a lot of unrelated entries. Let me know if I should have still updated it.

### DIFF
--- a/src/app/login/login.component.html
+++ b/src/app/login/login.component.html
@@ -23,7 +23,7 @@
           <img src="assets/runbox7_blue_dark.png" id="logoLogin" alt="Runbox 7" />
 	</div>
 	<mat-form-field>
-          <input matInput placeholder="Username" name="username" autofocus ngModel />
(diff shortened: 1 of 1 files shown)


tadzik:
Good call re the changelog – it is nowadays updated automatically right before deplyoment; we may as well clean it out of the repo (and the documentation).

Thank you for your contribution!

Issue #686, opened 2020-07-25
Default email address is not displayed as default.


davidbowdley:
I've just reloaded the app 3 times, and each time a different one of my aliases was selected as the default "from" address when I click Compose. Only once was it the correct address and the order of addresses listed in the drop down menu changed each time.

PR #730, opened 2020-08-19
fix(compose): Pick correct default email

kvviecien:
Fix to prevent picking a random From address while composing a new email.

Fixes: #686

### DIFF
--- a/src/app/compose/draftdesk.service.ts
+++ b/src/app/compose/draftdesk.service.ts
@@ -197,33 +197,45 @@ export class DraftDeskService {
     froms: FromAddress[] = [];
 
     constructor(public rmmapi: RunboxWebmailAPI) {
-        this.refreshDrafts().subscribe(() => {
+        this.refreshDrafts().then(() => {
             this.draftsRefreshed.next(true);
             this.draftsRefreshed.complete();
         });
     }
 
-    public refreshFroms(): Observable<FromAddress[]> {
-        return this.rmmapi.getFromAddress().pipe(
-            map(froms => this.froms = froms)
-        );
+    public async refreshFroms(): Promise<FromAddress[]> {
+        const froms = await this.rmmapi.getFromAddress().toPromise();
+        this.froms = froms.sort((a, b) => {
+            if (a.type === 'main') {
+                return -1;
+            } else if (b.type === 'main') {
+                return 1;
+            } else if (a.type === 'aliases') {
+                return -1;
+            } else if (b.type === 'aliases') {
+                return 1;
+            } else {
+                return 0;
+            }
+        });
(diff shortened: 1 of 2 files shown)


src/app/compose/compose.component.ts
```diff
@@ -109,18 +109,22 @@ export class ComposeComponent implements AfterViewInit, OnDestroy, OnInit {
...
+            
+            const defaultProfile = await this.rmmapi.getDefaultProfile().toPromise();
+            const defaultProfileIndex = await this.draftDeskservice.froms.findIndex(profile => profile.email === defaultProfile.email);
+
+            const from: FromAddress = this.draftDeskservice.froms.find((f) => 
                 f.nameAndAddress === this.model.from || f.email === this.model.from);
 
-            this.has_pasted_signature = false;
-            if (!from) {
+            if (!from || this.router.url.includes('compose?new=true')) {
```
tadzik, line 125: When is this second check needed?

kvviecien: Actually now it is. For some reason "this.model.from" isn't empty from the start so "!from" isn't enough, because there is always something inside "from". I believe it needs some more digging to see where in the app the model gets values assigned, but it's a temporary fix until I figure out what's going on there. 

Probably should've said all of that right from the start :D

castaway: I think you should be fixing the proper "from" here: https://github.com/runbox/runbox7/blob/master/src/app/compose/draftdesk.service.ts#L58

src/app/compose/draftdesk.service.ts
```diff
@@ -57,7 +57,6 @@ export class DraftFormModel {
 
     public static create(draftId: number, fromAddress: FromAddress, to: string, subject: string, preview?: string): DraftFormModel {
         const ret = new DraftFormModel();
-        ret.from = fromAddress.email;
```
castaway, line 60: Hm, I meant move the "pick from address" code to here, not remove it. This is the model, so I'd suggest do more work in the model, and less in the view

castaway: I understood the issue. My thought was do what you're doing, but do it in draftdesk.service instead, where its closer to the action, instead of the view

tadzik (approved):

src/app/compose/draftdesk.service.ts
kvviecien: Everything should be fixed now. Tadzik and I looked through it and just liked he described above - we sorted Froms, making the main identity win.

tadzik:
@castaway wanna have another look? :)

castaway (approved):
Good stuff!

PR #736, opened 2020-08-20
feat(webmail): add a way to save and reuse searches

tadzik:


tadzik:
@gtandersen sounds like a plan – I don't see any better option.

Updated now with a sync API. It's a bit simplistic and does no conflict resolution, but realistically the chance that someone will be adding saved searches from multiple offline devices at the same time is pretty slim, so we should be pretty safe.

tadzik:
Also, no need for a disable option I think – they'll disappear if none are stored.

src/app/app.component.html
```diff
@@ -166,6 +170,13 @@
               [searchInputField]="searchInputField"
               [currentFolder]="selectedFolder"
               matSuffix matTooltip="Show search options"></app-search-expression-builder> -->
+          <button *ngIf="searchText!==''"
+            mat-icon-button
+            matSuffix
+              (click)="showSaveSearchDialog()"
+              matTooltip="Save search">
+            <mat-icon>save</mat-icon>
```
castaway, line 178: svgIcons? gonna be a mess if we keep both kinds..

src/app/app.component.html
```diff
@@ -81,6 +81,10 @@
 
     <mat-divider></mat-divider>
 
+    <app-saved-searches (searchClicked)="searchFor($event)"></app-saved-searches>
```
castaway, line 84: This is the bit that reopens the search? Would it be possible to add more magic, so that if the search was saved with the extra search options (wrench) menu open, it opens + repopulates it?

tadzik: The fancy wrench options just modify the search query, so we do save it with everything in there. We could make it so that when a query is intered the wrench options automagically pick up what's in there, but that'll require pretty much reimplementing the xapian query parser :/

castaway: right, but we cant reload it and then edit it.. for eg just wanting to change the date on a saved one.. hmm could just also store a json rep of the fields in the advanced search?

tadzik: Storing JSON would work until someone modifies the query in the wrench UI and then adds some search terms in the regular search bar – already possible, and the changes are not reflected in the wrench UI then.

Sounds like a bigger task than what we have here – including the fact that our issue here is basically the symptom of the wrench UI not updating automatically after search queries are entered manually. I'll open a separate issue for this, as this will likely require exposing Xapian's query parser in our JS code if we wanna do it right (or plenty of regexes if we want to make it good enough).

tadzik: Moved to https://github.com/runbox/runbox7/issues/755 which links here, so resolving this for now :)

src/app/app.component.html
tadzik: Fixed in 3cef8ae, along with the other examples that sneaked in.

gtandersen (approved):
Brilliant @tadzik!

PR #737, opened 2020-08-21
Actually 632

tadzik:


castaway (approved):
Looks good to me (but I wrote it...)

PR #748, opened 2020-08-27
Fix compose edgecases

tadzik:
Fixes some edge cases related to delayed Froms and saving Drafts


castaway:
It looks like most of this one was already merged.

Issue #610, opened 2020-05-27
Select all messages button doesn't appear in wrapped list mode


davidbowdley:
Select all messages button doesn't appear on mobile portrait view.

gtandersen:
It also doesn't appear in wrapped list mode on any screen size.

PR #749, opened 2020-08-27
fix(canvastable): Show "select all messages" button in wrapped view

kvviecien:
Fixes: #610


tadzik (approved):
Hah, looks like it was intentional :) I wonder why.

PR #751, opened 2020-08-31
Folder count improvements

tadzik:


src/app/rmmapi/messagelist.service.ts
```diff
@@ -82,8 +82,8 @@ export class MessageListService {
                 if (msgFlagChange.flaggedFlag === true || msgFlagChange.flaggedFlag === false) {
                     this.messagesById[msgFlagChange.id].flaggedFlag = msgFlagChange.flaggedFlag;
                 }
-                // update both remote + local counts
-                this.refreshFolderList().then(folders => this.folderMessageCountSubject.next(this.getFolderCountsFor(folders)));
+                // update local results ASAP, schedule API results for later
+                this.refreshFolderCounts().then(() => this.refreshFolderList());
```
castaway, line 86: hm, isnt refreshFolderList the thing that pulls in the new data for counting? how does this work backwards? - just tested with sync off and yes.. the read count doesnt update when I mark some messages as "unread"

tadzik: I'd expect it work backwards since `refreshFolderList()` calls `refreshFolderCounts()` again once it's done updating `this.folders`.

castaway: ... nevermind I was using an out of date backend!

src/app/rmmapi/messagelist.service.ts
```diff
@@ -146,22 +145,15 @@ export class MessageListService {
                     }
 
                     this.folderListSubject.next(folders);
-                    // Moved from constructor (subscribing to our own
-                    // subjects seems silly)
-                    // Xapian counts:
+                    // Will fallback on the folder counters set above for folders not in the search index
                     this.refreshFolderCounts();
-                    // API counts:
                     this.folders.next(
```
castaway, line 150: can we keep the comments that help indicate which part happens where? this is the most confusing part of this code (some of it does things with indexes, some with REST APIs)

castaway (approved):
Looks good (cant comment on the xapian poking parts much..)

src/app/rmmapi/messagelist.service.ts
tadzik: Turns out nothing happened here, as this was leftover code :P Dropped it in 09186ed

PR #758, opened 2020-09-02
refactor(all): Retrieve MessageInfo and MailAddressInfo

castaway:
These went into runbox-searchindex, but are integral parts of rmm7

Also refactor message list/column stuff from app.component.ts into its own class/subclasses (messagelist, search, websocket)

Use these changes to set message rows/columns for each type of canvastable content.

Finally use it to select the message row correctly when reloading a url with a messageid in it.


gtandersen:
Well done @castaway.

Can you have a look @tadzik?

src/app/app.component.ts
```diff
@@ -1185,7 +1119,10 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
     if (this.singlemailviewer.messageId) {
       fragment += `:${this.singlemailviewer.messageId}`;
     }
-    this.router.navigate(['/'], { fragment });
+    // don't update if already there, its a bit circular!
+    if (!this.router.url.match('#fragment$')) {
```
tadzik, line 1123: Does this work? Looks like a weird string-regex hybrid.

castaway: Erm no.. I've just removed it as it should be redundant, now that the fragment subscribe code only saves the value, instead of doing stuff with it (it was at one point going around in circles)

tadzik (approved):
Nice work!

castaway:
local policy check isnt giving the same error travis is (tsk travis..)

PR #760, opened 2020-09-03
Startdesk experiments

tadzik:
This comments out all the mockup functionality for now: ideally they'd end up in separate components


tadzik:
All of this seems to have already been integrated elsewhere.

PR #765, opened 2020-09-09
fix(canvastable): Make sort icons show actual sorting direction

kvviecien:
Because of some typos, they weren't changing while sorting was changed, and only the "arrow up" icon was showing.

Additionally, moved the sorting icon to the right side of column name. This prevents the name from snapping to the side when the icon appears.


tadzik:
LGTM

tadzik (approved):

PR #768, opened 2020-09-15
fix(account): Don't require micro accounts to purchase addons for own domain usage

tadzik:
To be merged after the backend limitations are lifted.


gtandersen (approved):
Looks good @tadzik!

gtandersen (approved):

PR #770, opened 2020-09-18
test(mailviewer): Add a delay for reply/forwarding emails

kvviecien:
Delay needed for Travis CI to give it time to fully load content for tests.

### DIFF
--- a/e2e/cypress/integration/mailviewer.ts
+++ b/e2e/cypress/integration/mailviewer.ts
@@ -31,6 +31,7 @@ describe('Interacting with mailviewer', () => {
         cy.location().should((loc) => {
             expect(loc.pathname).to.eq('/compose');
         });
+        cy.wait(500);
         cy.get('mat-card-actions div').should('contain', "Re: No 'To', just 'CC'");
     });
 
@@ -41,6 +42,7 @@ describe('Interacting with mailviewer', () => {
         cy.location().should((loc) => {
             expect(loc.pathname).to.eq('/compose');
         });
+        cy.wait(500);
         cy.get('mat-card-actions div').should('contain', "Fwd: No 'To', just 'CC'");
     });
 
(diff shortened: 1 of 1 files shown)


e2e/cypress/integration/mailviewer.ts
```diff
@@ -31,6 +31,7 @@ describe('Interacting with mailviewer', () => {
         cy.location().should((loc) => {
             expect(loc.pathname).to.eq('/compose');
         });
+        cy.wait(5000);
```
tadzik, line 34: I think this'll have to happen earlier to be effective.

The bug happens (I think...) because our mailviewer is not ready in time for the fragment (`#Inbox:123`) parsing, so the delay will have to happen before that fragment is set. The only thing that comes to my mind is splitting the initial `cy.visit('/#Inbox:123')` into a `cy.visit('/'); cy.wait(for mailviewer to load, and only then...); cy.visit('/#Inbox:123');`.

Waiting where you put it happens after the reply button was clicked – and from what I gathered, the flapping test happens because sometimes there isn't even a reply button to click at that point – so maybe waiting for a bit right before that could also help? No idea. It'll take some Travis torment to determine ;)

kvviecien: I don't know for sure of course, but the error always returns `AssertionError: Timed out retrying: Expected to find element: 'mat-card-actions div', but never found it.`

`mat-card-actions div` is the title of currently composed email. From my understanding, it means that it already clicked the reply/forward button, and now it's looking for the email title. So if it can't find it, it means the compose view isn't fully loaded yet. 
That's why I put the timeout after the button click, to give time for the compose to load. 

The mailviewer loads without problems for any other tests so it has to be the compose not loading in time. 

5000ms is a bit too much I think, I believe I meant to put either 500 or 1000 there, so I'm going to change that.

gtandersen:
Thanks @kvviecien -- let's see if this works.

gtandersen (approved):

PR #778, opened 2020-09-30
test(mailviewer): Add delays for canvas to load

kvviecien:
Another fix for Travis. 
With those delays it should have enough time to load everything.


tadzik (approved):
Looks good, perhaps 1 second is a bit long, but it'll hardly be a chokepoint in e2e tests anyway.

PR #780, opened 2020-09-30
fix(account_security): dont display bottom pane "invalid password" msg

hernan604:


tadzik (approved):

src/app/rmm/account-security.ts
```diff
@@ -105,7 +105,6 @@ export class AccountSecurity {
           reply => {
             this.is_busy = false;
             if ( reply['status'] === 'error' ) {
```
tadzik, line 107: Looks like this entire if can now get removed, I think :)

PR #786, opened 2020-10-08
fix(calendar): Editing an item displays the correct times

### LINKED ISSUE #631: Calendar: date and time change when editing
When editing an event, the time and date is changed from the earlier defined to today+now.

castaway:
Fixes #631


tadzik (approved):

tadzik:
Thank you for fixing my nonsense ;)

Issue #745, opened 2020-08-27
Recurring calendar events switch days


davidbowdley:
Some recurring calendar events switch days after a period of time. In one case this is after two months for a weekly event.

castaway:
It would be handy if we had a known sample of this, or more details - was it created in runbox calendar or imported? I'm rewriting how recurrence is calculated (using ical.js to do all of it, rather than own code), so it would be good to have some known good/bad samples.

castaway:
Am inclined to close this as its from before I rewrote recurrence handling in the calendar

gtandersen:
@castaway Probably fine -- we can reopen it later if needed.

Issue #331, opened 2019-10-28
Add more recurring event options


davidbowdley:
Add more complex recurring options such as:

- [x] First day of the month.

- [ ] Second Wednesday of the month.

- [ ] Every other week/month.

- [ ] Weekly/monthly with ability to select more than one day of a week.

And other possibly combinations...

tadzik:
I think that "first day of the month" is actually covered if you create an event on the 1st and set it to recurring monthly. Is it currently not working as it should?

davidbowdley:
OK. Ticket that off for you now (of course that works, I am sure I was thinking of first something else probably).

castaway:
See recurrence notes on the RunboxCalendar page of our internal wiki.

PR #790, opened 2020-10-12
fix(calendar): Refactor ical event handling to treat each uid as one item

castaway:
Generate angular calendar event objects from "outside" the
RunboxCalendarEvent object (CalendarService now handles it), each RCE
object links to its main ICAL.Event object (which will be a separate
one if its an exception event).

TODO:
* [x] Tests! - Importing + see editing/saving notes
* [x] Sort out caching issue (caches fine, then re-caches at an extra level of depth!?) - FIXED
* Editing / saving of changes (changed nothing so far - needs to ensure we save/update recurring events, exceptions, or create new exceptions if needed?)
    * [x] 1. Can we edit a recurring event, and change (for eg) the time that it occurs, save updated ics back to DAV sanely 
    * [x] 2. Ditto above for non time related parts, eg description - DONE
    * [x] 3. Can we edit an event exception, change its time / description and save (whole event inc. exception) back to DAV
    * [x] 4. Can we create a new recurring event and save back to DAV (currently only daily/monthly etc, should clarify if thats "on same day/date of week/month" somehow on the UI) 
    * [x] 5. Can we edit a recurring event, change the time/ description on ONE of the instances, save back to DAV as a new exception
* [x] Ensure we handle/display recurring options we don't (yet) support gracefully - that is we can happily display events with "recurs 2nd Sunday of each month", but our UI doesn't yet support creating that type - grey out the "recurring" section of the edit dialog?
    * [x] Now implemented: editing of much more complex recurrence options: WEEKLY/MONTHLY/YEARLY, see images at https://gitlab.runbox.com/runbox/irc-meeting-actions/-/issues/274/ 
    * [x] Tests for the above needed!
* [x] Generate only the amount of events we can display, +/- 1 month - generating lots makes everything slllooow on events that never end (see Dave's iCloud examples)
    * [x] Update generated events when user moves to next/previous month, or next/previous week.. 
    * [x] Needs more visible "busy doing something" for event generation?  (via activities)
[x] Timezone support - set user timezone (from account settings) on new created events (and on floating/unset ones if edited) - display/edit things in user's timezone (no matter where their browser currently is)

* Simple test - 5 day event in April 2021, with one exception
[test45.txt](https://github.com/runbox/runbox7/files/5370279/test45.txt)

* Bigger test from iCloud/Dave, recurring daily/weekly/monthly with exceptions - no end dates! (slooww)
[RB Test2.txt](https://github.com/runbox/runbox7/files/5370287/RB.Test2.txt)

<s>See also matching backend branch. To use, set RUNBOX7_ANGULAR_BACKEND_HOST=http://rmm604test.gw01test.runbox.com:10004 before npm start.</s> - now live

Requires VTimezones on backend - https://gitlab.runbox.com/runbox/rmm/-/tree/castaway/vtimezones_for_ical

See also: #494 , #745, #331

### DIFF
--- a/src/app/calendar-app/calendar.service.ts
+++ b/src/app/calendar-app/calendar.service.ts
@@ -342,9 +342,6 @@ export class CalendarService implements OnDestroy {
         }
 
         return { 'id': id, 'event': vevents[0] };
-        // return this.generateEvents(undefined,
-        //                            undefined,
-        //                            [{ 'id': id, 'event': vevents[0] }]);
     }
 
     // generate RBE events from ICAL.Events (inc exceptions)
@@ -436,7 +433,6 @@ export class CalendarService implements OnDestroy {
 
     modifyEvent(event: RunboxCalendarEvent) {
         if (event._old_id) {
-            console.log('old id found');
             // special case: if event.calendar is being modified we can't simply update the event:
             // we need to copy it to a new calendar, and remove it from the old one.
             this.addEvent(event).then(id => {
@@ -494,17 +490,16 @@ export class CalendarService implements OnDestroy {
         console.log('Fetching events');
         this.activities.begin(Activity.RefreshingEvents);
         this.rmmapi.getCalendarEvents().subscribe(events => {
-            // Returns an array of RunboxCalendarEvent objects
             this.events = [];
             this.icalevents = [];
             events.forEach((e: any) => {
                 // store events into this.icalevents
                 this.importFromIcal(e.id, e.ical);
             });
-            console.log(this.icalevents);
 
             // generate RBE events for just this set, for current month+1:
-            // FIXME: what if user runs import while not looking at "today"?
+            // TODO: what if user runs import while not looking at "today"?
+            // Returns an array of RunboxCalendarEvent objects
             const runboxevents = this.generateEvents();
             this.events = this.events.concat(runboxevents);
             this.eventSubject.next(this.events);
@@ -551,37 +546,34 @@ export class CalendarService implements OnDestroy {
 
     // check if we have enough events to show another month worth
     // (forward and back), generate if not
-    updateEventList(viewPeriod?: ViewPeriod): Promise<void> {
-        return new Promise((resolve, reject) => {
-            this.activities.begin(Activity.GeneratingEvents);
-            let this_month, start, end_month, next_month: moment.Moment;
-            // Too much is better than not enough!
-            if (viewPeriod) {
-                this_month = moment(viewPeriod.start).startOf('month');
-            } else {
-                this_month  = moment().startOf('month');
-            }
-            // generate now(viewdate) +/- 1 month (max display size)
-            start       = this_month.clone(); start.subtract(1, 'month');
-            end_month   = this_month.clone(); end_month.add(1, 'month');
-            next_month  = this_month.clone(); next_month.add(3, 'month');
+    updateEventList(viewPeriod?: ViewPeriod) {
+        this.activities.begin(Activity.GeneratingEvents);
+        let this_month, start, end_month, next_month: moment.Moment;
+        // Too much is better than not enough!
+        if (viewPeriod) {
+            this_month = moment(viewPeriod.start).startOf('month');
+        } else {
+            this_month  = moment().startOf('month');
+        }
+        // generate now(viewdate) +/- 1 month (max display size)
+        start       = this_month.clone(); start.subtract(1, 'month');
+        end_month   = this_month.clone(); end_month.add(1, 'month');
+        next_month  = this_month.clone(); next_month.add(3, 'month');
 
-            // Could try and be clever and generate only last month / next month
-            // assuming some previous code did current one, but.. how to only insert
-            // those new ones into this.events?
-            // lets just try/do all for now
-            // we wont interrupt the page redraw process though.
+        // Could try and be clever and generate only last month / next month
+        // assuming some previous code did current one, but.. how to only insert
+        // those new ones into this.events?
+        // lets just try/do all for now
+        // we wont interrupt the page redraw process though.
 
-            // NB calendar-app.component.spec.ts relis on this being
-            // multiple months
-            if (this.icalevents.length > 0) {
-                this.events = this.generateEvents(start.toDate(), next_month.toDate(), this.icalevents);
-                this.eventSubject.next(this.events);
-            }
+        // NB calendar-app.component.spec.ts relis on this being
+        // multiple months
+        if (this.icalevents.length > 0) {
+            this.events = this.generateEvents(start.toDate(), next_month.toDate(), this.icalevents);
+            this.eventSubject.next(this.events);
+        }
 
-            this.activities.end(Activity.GeneratingEvents);
-            resolve();
-        });
+        this.activities.end(Activity.GeneratingEvents);
     }
 
 }
--- a/src/app/calendar-app/calendar.service.ts
+++ b/src/app/calendar-app/calendar.service.ts
@@ -519,8 +519,6 @@ export class CalendarService implements OnDestroy {
             // multiple months
             if (this.icalevents.length > 0) {
                 this.events = this.generateEvents(start.toDate(), next_month.toDate(), this.icalevents);
-//                this.events = this.generateEvents(viewPeriod.start, viewPeriod.end, this.icalevents);
-                // calendar-app calls filterEvents, will this recurse infinitely!?
                 this.eventSubject.next(this.events);
             }
 
--- a/src/app/calendar-app/calendar.service.ts
+++ b/src/app/calendar-app/calendar.service.ts
@@ -71,7 +71,7 @@ export class CalendarService implements OnDestroy {
             }
             console.log('Cache version:', cache['version']);
             // tslint:disable-next-line:triple-equals
-            if (cache['version'] != 3) {
+            if (cache['version'] != 4) {
                 console.log('Old cache format found, removing');
                 storage.set('caldavCache', undefined);
                 return;
@@ -87,7 +87,9 @@ export class CalendarService implements OnDestroy {
             this.calendarSubject.next(this.calendars);
             this.eventSubject.next(this.events);
             this.activities.end(Activity.LoadingCache);
-        }).then((_) => this.syncCaldav());
+        }).then(
+            () => this.syncCaldav(),
+        );
 
         this.calendarSubject.subscribe(cals => {
             const updatedCals = [];
@@ -202,7 +204,7 @@ export class CalendarService implements OnDestroy {
         // fixup underspecified rules
         // (or else figure out the nearest one to our start date but ick)
         // clone the recur object to update it:
-        let new_rule = rule_iterator.rule.clone();
+        const new_rule = rule_iterator.rule.clone();
         const event_start = icalevent.startDate;
         if (rule_iterator.rule.freq === 'WEEKLY'
             && !rule_iterator.has_by_data('BYDAY')) {
@@ -249,8 +251,6 @@ export class CalendarService implements OnDestroy {
         }
         // update if changed:
         if (new_rule.toString() !== rule_iterator.rule.toString()) {
-            console.log('new RRULE');
-            console.log(new_rule.toString());
             icalevent.component.updatePropertyWithValue('rrule', new_rule);
         }
 
@@ -360,7 +360,6 @@ export class CalendarService implements OnDestroy {
 
             // improve RRULEs if underspecified (then save!?)
             icalevent = this.fixICALEvent(icalevent);
-            // FIXME: splice changed item back into .icalevents? (in fixICALevent func)
 
             // Now recheck the start_date_ICAL against the known
             // start/end/until dates
@@ -372,7 +371,7 @@ export class CalendarService implements OnDestroy {
                 const calc_end_date = ICAL.Time.fromString(icalevent.component.parent
                     .getFirstPropertyValue('X-RUNBOX-CALCULATED-DTEND')).toJSDate();
                 if (startDate && moment(startDate).isAfter(calc_end_date)) {
-                    console.log('Skipping generation as already ended');
+                    // console.log('Skipping generation as already ended');
                     return;
                 }
                 // on COUNT, generate all
--- a/src/app/calendar-app/calendar.service.ts
+++ b/src/app/calendar-app/calendar.service.ts
@@ -22,12 +22,14 @@ import { RunboxCalendarEvent } from './runbox-calendar-event';
 import { RunboxWebmailAPI } from '../rmmapi/rbwebmail';
 import { StorageService } from '../storage.service';
 import { BackgroundActivityService } from '../common/background-activity.service';
+import { ViewPeriod } from 'calendar-utils';
 
 import { Injectable, OnDestroy } from '@angular/core';
 import { HttpErrorResponse } from '@angular/common/http';
 import { Observable, Subject, ReplaySubject } from 'rxjs';
 
 import * as moment from 'moment';
+import * as ICAL from 'ical.js';
 
 enum Activity {
     LoadingCache        = 'Loading local cache',
@@ -45,6 +47,8 @@ enum Activity {
 export class CalendarService implements OnDestroy {
     calendars:    RunboxCalendar[]      = [];
     events:       RunboxCalendarEvent[] = [];
+    // Stores ICAL.Events, one per single/recurring item, with .exceptions set
+    icalevents:   { id: string, event: ICAL.Event}[] = [];
 
     syncTokens = {};
     syncInterval: any;
@@ -78,11 +82,12 @@ export class CalendarService implements OnDestroy {
             for (const cal of this.calendars) {
                 this.syncTokens[cal.id] = cal.syncToken;
             }
+            const runboxevents = cache['events'].map((e: any) => this.fromIcal(e.id, e.ical));
+            runboxevents.forEach((events) => this.events = this.events.concat(events));
             this.calendarSubject.next(this.calendars);
-            this.events = cache['events'].map((e: any) => new RunboxCalendarEvent(e.id, e.jcal));
             this.eventSubject.next(this.events);
             this.activities.end(Activity.LoadingCache);
-        });
+        }).then((_) => this.syncCaldav());
 
         this.calendarSubject.subscribe(cals => {
             const updatedCals = [];
@@ -110,7 +115,6 @@ export class CalendarService implements OnDestroy {
         this.syncInterval = setInterval(() => {
             this.syncCaldav();
         }, this.syncIntervalSeconds * 1000);
-        this.syncCaldav();
     }
 
     ngOnDestroy() {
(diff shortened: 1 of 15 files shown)


src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +88,437 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
     }
 
+    // If set on an event with RRULE, updates entire repeated event
+    // to now *start* from this new date
+    // If set on an exception, only changes that exception
     set dtstart(value: moment.Moment) {
         const allDay = this.allDay;
-        this.event.startDate = ICAL.Time.fromJSDate(value.toDate());
+        this._dtstart = this.momentToIcalTime(value, this.event.startDate ? this.event.startDate.zone : null);
+        // FIXME: Do this only if its the first one?
```
tadzik, line 103: I don't think I'm getting these two FIXMEs. How does editing a single event, down the recurrence chain, impact all the others from here?

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +88,437 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
-            ? this.icalTimeToLocalMoment(this.event.endDate)
+        return this._dtend
+            ? this.icalTimeToLocalMoment(this._dtend)
             : undefined;
     }
 
+    // If set on an event with RRULE, updates entire repeated event
+    // ?? Does this shorten it to one day ? (does it include the date)
+    // If set on an exception, only changes that exception
+    // FIXME: Does this change an ICAL.Time with no date (isDate=true) to
```
tadzik, line 119: Good question – I guess so, since Moment doesn't really have such a distinction, and it's impossible to infer what the user meant from the datetime alone. We may need to spread ICAL.Time further around our code to fix this :|

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +88,437 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
 
     get allDay(): boolean {
         // isDate in ICAL.Event means "has no time"
-        return this.event.startDate.isDate;
+        return this._dtstart.isDate;
+    }
+
+    get recurs(): boolean {
+        const recur = this.event.component.getFirstPropertyValue('rrule');
+        return recur ? true : false;
```
tadzik, line 160: Seems like we could just skip this variable entirely :)

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +88,437 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
     }
 
+    // This is the original DTSTART of the event - if recurring, will
+    // be different from dtstart above.
+    get recurStart(): Date {
+        return this.event.startDate.toJSDate();
+    }
+
+    set recurStart(start: Date) {
+        // Nope can't do that Dave
```
tadzik, line 181: Maybe log the attempt anyway? Unless this will never happen due to a user action.

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +88,437 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
+
+    // replace entire list of day(s)
+    set recursByDay(value: string[]) {
+        const recur = this.event.component.getFirstPropertyValue('rrule');
+        recur.setComponent('byday', value);
+        this.event.component.updatePropertyWithValue('rrule', recur);
+    }
+
+    get recursByMonthDay(): string[] {
+        const recur = this.event.component.getFirstPropertyValue('rrule');
```
tadzik, line 273: Maybe these could get refactor into a private get/set `recursBy(what)` to avoid repetition?

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +88,437 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
+        this._dtstart.isDate = value;
+        if (this._dtend) {
+            this._dtend.isDate = value;
         }
         // I know this looks silly, but without this
         // ICAL.Event will not notice the isDate change
         // and the event metadata will still be off
-        this.event.startDate = this.event.startDate;
-        this.event.endDate   = this.event.endDate;
+        // FIXME: where is this set/used?
```
tadzik, line 326: Hm, perhaps nowhere anymore? If the tests are still passing without it...

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +88,437 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
-                [ 'vevent',
-                    [
-                        [ 'dtend',   {}, 'date-time', '2019-10-02T14:00:00' ],
-                        [ 'dtstart', {}, 'date-time', '2019-10-02T12:00:00' ],
-                        [ 'summary', {}, 'text', '' ] ],
-                    []
-                ]
-            ] ]
+            new ICAL.Event(new ICAL.Component('vevent')),
+            ICAL.Time.fromData({year: 2019, month: 10, day: 2, hour: 14, minute: 0}),
```
tadzik, line 340: Maybe we can take this opportunity to actually follow the comment and create that even for today ;)

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +88,437 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
+            new_exception.summary = title;
+        }
+        if (description) {
+            new_exception.description = description;
+        }
+        if (location) {
+            new_exception.location = location;
+        }
+        this.ical.addSubcomponent(new_exception.component);
+        // regenerate this.event to sort out the exceptions/recurrences?
```
tadzik, line 421: Looks like this always returns false: perhaps it doesn't need a return type at all?

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +88,437 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
+        recur_months:   string[],
+        recur_nth:      string[],
+    ) {
+        if (this.recurs && recur_save_type !== '0') {
+            // Adding an exception to an already recurring event
+            this.addExceptionEvent(
+                this.dtstart,
+                dtstart,
+                dtend,
+                recur_save_type === '2' ? true : false,
```
tadzik, line 447: Bits like these (and related in the editor dialog) make me think that this should be an enum, not a string :)

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -258,11 +561,12 @@ export class RunboxCalendarEvent implements CalendarEvent {
             id:       this.id,
             calendar: this.calendar,
             jcal:     this.ical.toJSON(),
+            ical:     this.toIcal(),
```
tadzik, line 564: Wrong place for this comment (Github won't let me add one to unchanged lines), but we should probably remove the `// returns jCal` comment from this since it hasn't been correct for a while, from the looks of it :) Actually, I wonder: do we really need both `jcal` and `ical` in there?

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -273,4 +577,14 @@ export class RunboxCalendarEvent implements CalendarEvent {
             return moment.utc(utc.toString()).tz(moment.tz.guess());
         }
     }
+
+    private momentToIcalTime(input: moment.Moment, zone: ICAL.Timezone): ICAL.Time {
+        if (!zone) {
+            // This should probably use tz.guess() ?
```
tadzik, line 583: Probably, like the method above.

src/styles.scss
```diff
@@ -150,6 +150,10 @@ mat-chip.mat-chip.mat-standard-chip {
 
 /* Generic custom styles */
 
+.mat-dialog-container {
```
tadzik, line 153: Not sure if this particular bit is to blame, but just in case: somewhere in this PR the event editors in mobile view became narrow enough in mobile view to produce an extra scrollbar, making it a bit awkward: 

https://user-images.githubusercontent.com/86378/106602245-05498400-655d-11eb-893c-a22e2f59530b.mp4

tadzik (changes requested):
Looks great overall: it almost feels like it would've been easier to rewrite it than fix it ;)

Found some nitpicks here and there, and two new minor visual bugs, documented with videos.

src/app/calendar-app/runbox-calendar-event.ts
castaway: _dtstart (and dtstart and start) are all "start date/time of this event instance" .. event.startDate is the start of whole recurring event, if it is one.. so we shouldnt be changing that unless the object we're editing is the original start event.

I think I subsequently prevented this by disallowing "All occurences" if we were editing a non-first repeat event. (I'll add another check here just in case)

src/app/calendar-app/runbox-calendar-event.ts
castaway: I am using it in quite a few places..

src/app/calendar-app/runbox-calendar-event.ts
castaway: Nothing in this code attempts to set it - does ts have the concept of a read-only attribute ?

tadzik: I think that's what happens when you specify a `get` only.

src/app/calendar-app/runbox-calendar-event.ts
castaway: I thought of that but they're not all the same, I guess a few don't have special cases..

src/app/calendar-app/runbox-calendar-event.ts
castaway: Crazy idea.. ;)
Luckily ICAL.Time has a function for that..

src/app/calendar-app/runbox-calendar-event.ts
castaway: Nope, and since I updated the cache version anyway, might as well remove it

castaway: .. in fact we're not using it at all, the saveCache does 

```({'id': event['id'], 'ical': event['event'].component.parent.toString() } )```

File: src/styles.scss (hunk 12 above)
Comment (castaway) [reply src/styles.scss:153]: @kvviecien thoughts?

Comment (kvviecien) [reply src/styles.scss:153]: I'll try to figure out what's happening there and fix it. Expect a pr!

**Comment by castaway** [comment]
Just found another while testing against prod - the sabre database has some old/junk in it - from the days when an RRULE and its EXCEPTION events used to be stored separately - my new parsing can't fathom how to cope with an EXCEPTION on its own.. :(

File: src/app/calendar-app/runbox-calendar-event.ts (hunk 8 above)
Comment (castaway) [reply src/app/calendar-app/runbox-calendar-event.ts:415]: you mean addExceptionEvent ? it doesn't have one *confused*

Comment (tadzik) [reply src/app/calendar-app/runbox-calendar-event.ts:415]: It does on my screen :)

```
    addExceptionEvent(
        origdate: moment.Moment,
        startdate: moment.Moment,
        enddate: moment.Moment,
        thisandfuture: boolean,
        title: string,
        description: string,
        location: string): boolean {
        if (this.isException) {
            // This shouldnt be possible
            console.log('Refusing to create an exception of an exception');
            return false;
        }
```

File: src/app/calendar-app/runbox-calendar-event.ts (hunk 9 above)
Comment (castaway) [reply src/app/calendar-app/runbox-calendar-event.ts]: Yup agreed, fixed

File: src/app/calendar-app/runbox-calendar-event.ts (hunk 8 above)
Comment (castaway) [reply src/app/calendar-app/runbox-calendar-event.ts:421]: Ah! Oops

File: src/styles.scss (hunk 12 above)
Comment (kvviecien) [reply src/styles.scss:153]: So I didn't... I played with it for hours and I have no idea why the scroll appears there.
I'll try to rewrite some of it though. I think the HTML could be faulty, a lot's happening there.

File: src/app/calendar-app/runbox-calendar-event.ts (hunk 11 above)
Comment (castaway) [reply src/app/calendar-app/runbox-calendar-event.ts:573]: This shaves a whole new yak, as if we guess and find user's timezone (or grab the one configured in the account), we now need to have a VTIMEZONE entry to store with it. I'm now looking into https://github.com/libical/vzic as a source of these.

File: src/app/calendar-app/calendar.service.ts [hunk 13]
```diff
@@ -204,19 +406,19 @@ export class CalendarService implements OnDestroy {
...
             this.addEvent(event).then(id => {
                 console.log('Event recreated as', id);
-                this.deleteEvent(existing.id);
+                this.deleteEvent(event._old_id);
             });
         } else {
             // simple case: just modify the event in place
             this.activities.begin(Activity.EditingEvent);
             this.rmmapi.modifyCalendarEvent(event as RunboxCalendarEvent).subscribe(_ => {
+                // FIXME: ids are no longer unique!? add start time to the mix?
```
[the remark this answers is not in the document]
castaway: Figured it out - was me breaking the "delete/re-add if we're changing calendars" test

src/app/calendar-app/runbox-calendar-event.ts
castaway: Removed fixme and verified working correctly

src/app/calendar-app/runbox-calendar-event.ts
castaway: Huh I thought it didnt compile without - seems it does, so I'll remove it

src/app/calendar-app/runbox-calendar-event.ts
castaway: Sorted - that was fun..

src/app/calendar-app/runbox-calendar-event.ts
castaway: We have tests using both dates and datetimes, so hopefully we're good..

castaway:
There's still one e2e (calendar) test that's complaining about hidden info it can't "see" :(

src/app/calendar-app/calendar.service.ts
```diff
@@ -204,19 +435,19 @@ export class CalendarService implements OnDestroy {
     }
 
     modifyEvent(event: RunboxCalendarEvent) {
-        const existing = this.events.find(c => c.id === event.id);
-        if (existing.calendar !== event.calendar) {
+        if (event._old_id) {
+            console.log('old id found');
```
tadzik, line 439: Looks like a leftover, this log

src/app/calendar-app/calendar.service.ts
```diff
@@ -237,11 +468,45 @@ export class CalendarService implements OnDestroy {
...
         this.rmmapi.getCalendarEvents().subscribe(events => {
-            this.events = events.map((e: any) => RunboxCalendarEvent.fromIcal(e.id, e.ical));
+            // Returns an array of RunboxCalendarEvent objects
+            this.events = [];
+            this.icalevents = [];
+            events.forEach((e: any) => {
+                // store events into this.icalevents
+                this.importFromIcal(e.id, e.ical);
+            });
+            console.log(this.icalevents);
```
tadzik, line 504: Leftover :)

src/app/calendar-app/calendar.service.ts
```diff
@@ -237,11 +468,45 @@ export class CalendarService implements OnDestroy {
...
+            }
+        });
+    }
+
     reloadEvents() {
         console.log('Fetching events');
         this.activities.begin(Activity.RefreshingEvents);
         this.rmmapi.getCalendarEvents().subscribe(events => {
-            this.events = events.map((e: any) => RunboxCalendarEvent.fromIcal(e.id, e.ical));
+            // Returns an array of RunboxCalendarEvent objects
```
tadzik, line 497: What does?

src/app/calendar-app/calendar.service.ts
```diff
@@ -282,4 +548,40 @@ export class CalendarService implements OnDestroy {
             this.activities.end(Activity.RefreshingCalendars);
         });
     }
+
+    // check if we have enough events to show another month worth
+    // (forward and back), generate if not
+    updateEventList(viewPeriod?: ViewPeriod): Promise<void> {
```
tadzik, line 554: Hmm, any reason why this returns a Promise? Seems completely synchronous to me.

src/app/calendar-app/event-editor-dialog.component.ts
```diff
@@ -60,36 +104,253 @@ export class EventEditorDialogComponent {
         this.calendars = data['calendars'];
         this.calendarFC.setValue(this.calendars[0].id);
 
-        if (data['event']) {
-            this.event = data['event'];
+        this.event = data['event'];
+        if (!data['is_new']) {
             this.calendarFC.setValue(this.event.calendar);
             this.export_url = '/rest/v1/calendar/ics/' + this.event.id;
```
tadzik, line 110: For later: the specifics of this URL may be better off being abstracted away by RunboxWebmailAPI

src/app/calendar-app/event-editor-dialog.component.ts
```diff
@@ -60,36 +104,253 @@ export class EventEditorDialogComponent {
...
+            this.event_title = this.event.title;
+            this.event_location = this.event.location;
+            this.event_description = this.event.description;
+            this.event_allDay = this.event.allDay;
+
+            // this.event_start = this.event.dtstart.toDate();
```
tadzik, line 116: Leftover :)

src/app/calendar-app/event-editor-dialog.component.ts
```diff
@@ -111,24 +372,40 @@ export class EventEditorDialogComponent {
             // so let's help it a little
             this.calendarFC.markAsTouched();
             return;
-        } else {
-            this.event.calendar = this.calendarFC.value;
         }
+        // else if (this.event.calendar !== this.calendarFC.value) {
```
tadzik, line 376: Leftover :)

src/app/calendar-app/runbox-calendar-event.spec.ts
```diff
@@ -40,8 +51,244 @@ describe('RunboxCalendarEvent', () => {
...
+            true,
+            sut.recurringFrequency,
+            sut.recurInterval,
+            undefined, undefined, undefined, // and optional params..
+        );
+
+        expect(sut.toIcal()).toContain('SUMMARY:Moved weekly event');
+        expect(sut.toIcal()).toContain('RECURRENCE-ID;VALUE=DATE:20210201');
+        expect(sut.toIcal()).toContain('DTSTART;VALUE=DATE:20210204');
+        // expect(events.length).toBe(2);
```
tadzik, line 174: Are those a TODO to be fixed, or a leftover?

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +98,434 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
         this.allDay = allDay;
     }
 
     // angular-calendar compatibility
 
     get start(): Date {
-        return this.dtstart.toDate();
+        // _dtstart as-is for display (its already got tz set.. )
+        return new Date(this._dtstart.toString());
+        // this one converts to utc which is not helpful.
```
tadzik, line 146: We can probably remove that bit then :)

src/app/calendar-app/runbox-calendar-event.ts
```diff
@@ -69,137 +98,434 @@ export class RunboxCalendarEvent implements CalendarEvent {
...
         if (this.allDay) {
-            shownEnd.subtract(1, 'days');
+            shownEnd.addDuration(new ICAL.Duration({'isNegative': true, 'days': 1}));
         } else {
-            shownEnd.subtract(1, 'seconds');
+            shownEnd.addDuration(new ICAL.Duration({'isNegative': true, 'seconds': 1}));
         }
-        return shownEnd.toDate();
+        return new Date(shownEnd.toString());
+        // return shownEnd.toJSDate();
```
tadzik, line 163: Can probably go too.

src/app/calendar-app/runbox-calendar-event.ts
tadzik: Ah, I was confused and assumed `getFirstProperlyVaule` returns a boolean which we could just return. We could swap this whole body for something like `return !!this.event.component.getFirstPropertyValue('rrule');`, though that doesn't really make it any more readable :)

src/app/calendar-app/runbox-calendar-event.ts
tadzik: Right, they are subtly distinct now that I've paid more attention to it.

tadzik (commented):
I think I only found some confusing commented out code – looks pretty much ready to go :)

src/app/calendar-app/calendar.service.ts
castaway: Erm, dunno? cargo-culting probably!

src/app/calendar-app/runbox-calendar-event.spec.ts
castaway: Leftover *removes*

castaway:
> There's still one e2e (calendar) test that's complaining about hidden info it can't "see" :(

"fixed" it by removing the "set Calendar" part of the test (we only have one mock calendar and thats the default)

Test now runs fine if run on its own, sometimes has issues when run with everything else (gets one of our oft seen errors..)

tadzik (approved):

gtandersen (approved):
Well done @castaway and @tadzik!

Issue #574, opened 2020-05-13
Contacts: Support preferred/primary email/tel


castaway:
The vCard spec allows the storage of a preferred/primary email or telephone number, when multiple email addresses are supplied. Could we:

1) Flag/Checkbox in the Email/Tel list for a contact which entry is the preferred one
2) Import/export this data via vcf
3) (for bonus points), pick the preferred email by default when a contact is selected on the "compose" screen

PR #796, opened 2020-10-22
fix(contacts, calendar): Warn user on incorrect import file types

### LINKED ISSUE #583: Importing a non-ICS into Calendar results in a broken dialog
We don't indicate the format of the import input, and we don't handle anything other than proper .ics correctly.

castaway:
Fixes #583, #574


gtandersen (approved):
Great @castaway -- approving this in anticipation of passing tests. :)

PR #798, opened 2020-10-27
Fix bogus carts

tadzik:
People are ending up with products not available for purchase in their shopping carts, for some reason. A backend fix is in the pipeline to hopefully eliminate these situations, but this should at least enable people to purchase the working things, without locking the cart with products that can't even be removed from it easily.


gtandersen (approved):

tadzik:
(No failures locally, just Travis being annoying – can someone verify?)

PR #800, opened 2020-10-28
fix(webmail): Fix URL navigation

tadzik:
Drafting this as it appears to break some tests (and doesn't pass any new ones yet :))


tadzik:
Now contains a test that does not pass on current master, confirming that it actually fixes something :)

castaway (approved):
I think this looks fine, would be handy if we could have a test for verifying that the correct row is selected on visit to #Inbox:XX (eg highlighted correctly) - admittedly I should have added one!

Assuming we're still calling selectRowByMessageId when the fragment changes, we should be good

Issue #424, opened 2020-01-11
Restructure Account Products


davidbowdley:
This section also shows expired products that are not "active" (which is useful and desirable) but the title of the section may lead to a misunderstanding.

tadzik:
That'd make a bit confusing since https://runbox.com/app/account/components and its menu item is already called “Account Products”. We could rename it to “Account Components”, since that's how we refer to it in the code.

davidbowdley:
I that case  I would call it "Main Account Products" and move the first section of products that are in "Account add-ons" to this section as they only relate to the Main account in the sense that the Main account holder manages them. I would then change "Account add-ons" to "Sub-accounts".

gtandersen:
The reason it's called **Active products** is that it lists all active/existing products on the main account, whether they are sub-accounts or other add-ons.

**Account add-ons** lists all _available_ products, not just those already subscribed to by the main account.

To make this clearer we could split the page in two sections, first listing existing/current details and then available add-ons.

We could also change "Account upgrades" to "Account plans" or similar since technically you can both up- and downgrade your subscription plan.

So, to hopefully be consistent and better organized we could then end up with:

### Your Products
```
| Your Account Plan | Your Active Products | Your Transactions |
```
### Available Add-ons
```
| Main Account Add-ons | Sub-Account Products |
```

davidbowdley:
"Active products" also lists inactive products which is why I suggested the change.

gtandersen:
@davidbowdley Then we can rename **Your Active Products** to just **Your Products**.

Do we agree on the other proposed changes in my comment?

davidbowdley:
@gtandersen Yes, we could do it that way and it would be a big improvement on what we have now. 

The only further suggestion is to call Your Account Plan something different. My suggestions are Main Account Plan, Your Main Account Plan or Your Main Account. Of all those I would probably choose the last one to retain "Your" and add "Main" and still keep it to three words.

The reason being is that terminology is crucial as we've discussed with other services we offer and if we can hammer home the idea of Main and Sub-accounts it makes it easier for customers to understand what we say elsewhere and when they contact support.

gtandersen:
@davidbowdley In that case I think we should call it **Your Main Account Plan** since it's then consistent with **Your Products** and **Your Transactions** --  and I think it's useful to include "Plan" here because it's a common term used for this type of product.

[...]

PR #808, opened 2020-11-06
feat(account): Add a way to access account settings on mobile

tadzik:
This introduces a way to access the account screen from the mobile sidemenu.

Looks like this:

test

Imitating (visually and in behaviour) the same sidemenu as in webmail, contacts etc. This will enable access to all the account-related components and allow the user to switch between them. Before this is rolled out we need to ensure all of these components are actually accesible on mobile, and fix the ones that are not either in this PR or some others.

We can also take the opportunity to rename some of the sections to make them more understandable – something that @davidbowdley has often brought up.

Known issues:

- [x] The main "Account settings" page – we could have it be a hub between all the account-related pages – kind of what /account/components currently is, but not only for products but also for account security, identities and all the other things that are currently inaccessible on mobile. It could potentially just replace /components completely.
- [x] Identities and account security do not have the sidemenu visible in them. We could implement it in them, or include their main components (with the existing `<mat-sidenav>` wrappers removed) in the existing account-app architecture (components "injected" into account-app.component.html using subpaths. (https://github.com/runbox/runbox7/issues/817 and https://github.com/runbox/runbox7/issues/818)
- [x] Identities screen is not mobile friendly (widths hardcoded to too wide values) (https://github.com/runbox/runbox7/issues/819)
- [ ] Account security is not mobile friendly (some elements overlap or become too narrow in tables) (https://github.com/runbox/runbox7/issues/820)
- [x] The special offer on subscriptions page is not mobile friendly (button overlaps text) (https://github.com/runbox/runbox7/issues/821)
- [x] Addons cards are too wide and don't scale down on mobile screens (https://github.com/runbox/runbox7/issues/826)
- [x] Active Products table becomes too narrow to be readable on mobile *and* doesn't fit on the screen
- [x] Transactions table becomes too narrow to be readable on mobile *and* doesn't fit on the screen (https://github.com/runbox/runbox7/issues/827)
- [x] The shopping cart icon overlaps some page headers (including the shopping cart page itself). Maybe it should be moved to the toolbar, regardless of whether we're on mobile or not? (https://github.com/runbox/runbox7/issues/828)
- [x] The shopping cart table is too narrow to be readable (https://github.com/runbox/runbox7/issues/829)
- [x] The payment method cards are too big, overlap and don't fit on the screen (https://github.com/runbox/runbox7/issues/830)
- [x] The payment form(s) don't fit on mobile screens (https://github.com/runbox/runbox7/issues/831)


gtandersen (approved):
Looks great @tadzik, and I agree with your notes.

See also #424 regarding restructuring Account Products.

tadzik:
3679dc8 implements a mobile-friendly renewals page, making the table narrower but with expandable rows:

active

The desktop version remains as it was.

gtandersen:
Looks beautiful, @tadzik!

tadzik:
The entire payment flow is now "mobilized" (with some nice-to-have tuneups pending in https://github.com/runbox/runbox7/issues/831). @kvviecien do I remember correctly that you were working on the identities, or was it something else that's in progress?

kvviecien:
> The entire payment flow is now "mobilized" (with some nice-to-have tuneups pending in #831). @kvviecien do I remember correctly that you were working on the identities, or was it something else that's in progress?

That's right, I'm on identities, so you can grab another one :)

Issue #425, opened 2020-01-11
Sub-accounts no longer listed next to sub-account products


davidbowdley:
This relates to the listing of current/active products in Runbox 7.

In Runbox 6 it was possible to see a pop-up list of sub-accounts that are associated with a sub-account product. This was useful so customers could check the product they were going to renew was the one they intended to renew.

This isn't possible in Runbox 7 and would be incredibly useful.

tadzik:
That'd make a bit confusing since https://runbox.com/app/account/components and its menu item is already called “Account Products”. We could rename it to “Account Components”, since that's how we refer to it in the code.

PR #811, opened 2020-11-09
feat(account): Add a way to view subaccounts associated with a product

tadzik:
Fixes GH-425.

Looks like this:

Hovered: 
image

Clicked:
subs2


gtandersen (approved):
Looks great @tadzik!

Suggestion (perhaps for later):  It might be easier to read the numbers if they had dedicated columns, e.g. Available and Used.

PR #825, opened 2020-11-17
feat(sentry): Include user data in error reports

tadzik:
Couldn't figure out an easier way to do it without breaking RunboxWebmailAPI (see the comments), so it is a bit awkward.


tadzik:
Maybe instead we could have RunboxWebmailAPI just call sentry with user details once it obtains them? Not sure if that's actually more elegant though.

gtandersen (approved):

castaway:
This seems OK to me, it feels like its better for sentry code to use the runbox code from an external perspective, rather than tangle sentry+runbox api by having the API tell sentry when its done getting a user..

PR #833, opened 2020-11-23
feat(usage): Add usage stats for tracking the popularity of components/settings

tadzik:


gtandersen (approved):
Great, @tadzik!

src/app/app.component.ts
```diff
@@ -167,6 +168,7 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
     private hotkeysService: HotkeysService,
     public settingsService: AppSettingsService,
     private savedSearchService: SavedSearchesService,
+    _usage: UsageReportsService,
```
castaway, line 171: this doesnt get set / used?

castaway (changes requested):
Bit confused about _usage in app.component.ts .. rest seems sane!

src/app/app.component.ts
tadzik: It is not used anywhere, but putting it here makes Angular instantiate it the moment we open the webmail for the first time (but no sooner), so that we can start periodically checking webmail's settings in there. I couldn't think of any more elegant way to do it – any ideas?

castaway: Umm, nope other than adding a comment for our future selves

castaway (approved):
Works for me!

Issue #834, opened 2020-11-24
Import of large VCF file (500+ contacts) - errors out on some


castaway:
Weirdly this works on dev (although it does take a good number of minutes), on prod we get timeouts on some of the "saveContact" calls (we are parsing the file and then calling saveContact for each individual entry)

We also then (at the same time) run a contacts sync (500x times), once per item in the file

gtandersen:
@castaway Is this still occurring in prod?

castaway:
Haven't tried it myself, but I imagine so..

PR #835, opened 2020-11-24
fix(contacts): Only sync once during import of many contacts

castaway:
Attempting to run all contact imports, then do the reload() rather than one per contact (which is a bit crazy for a large import)


castaway:
Should assist with, but may not actually fix, #834

tadzik (approved):

PR #837, opened 2020-11-25
style(start): Improve responsivity for mobile screens

kvviecien:
Mobile view fixed. 
Also commented out the "panel controls" from the view because they had no function yet.

<kbd></kbd>


gtandersen (approved):
Looks great @kvviecien!

tadzik:
Looks good :)

I wonder if in mobile view a dropdown `<select>` wouldn't be easier to use than the MaterialTabs we're currently using – though that's probably a more involved change so we can try this later if needed.

PR #838, opened 2020-11-26
fix(start): Fix case sensitivity for address matching

kvviecien:
If the email we received was sent to us with some capitalized letters in the address i.e. `ABC@runbox.com`, the overview didn't know it's our address and was showing it instead of the sender's email address.

### DIFF
--- a/src/app/start/startdesk.component.ts
+++ b/src/app/start/startdesk.component.ts
@@ -75,7 +75,7 @@ export class StartDeskComponent implements OnInit {
 
     ngOnInit() {
         this.rmmapi.getFromAddress().subscribe(
-            froms => this.ownAddresses.next(new Set(froms.map(f => f.email))),
+            froms => this.ownAddresses.next(new Set(froms.map(f => f.email.toLowerCase()))),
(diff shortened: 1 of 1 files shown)


tadzik:
Are we `.toLowerCase`-ing ownAddresses as well when it's being loaded? Not sure where these come from, but it'd probably a good idea to lowercase these too.

src/app/start/startdesk.component.ts
```diff
@@ -75,7 +75,7 @@ export class StartDeskComponent implements OnInit {
 
     ngOnInit() {
         this.rmmapi.getFromAddress().subscribe(
-            froms => this.ownAddresses.next(new Set(froms.map(f => f.email))),
+            froms => this.ownAddresses.next(new Set(froms.map(f => f.email.toLowerCase()))),
```
kvviecien, line 78: I `.toLowerCase`-d the froms in ownAddresses while they're being loaded for this component as well now. Could save some issues with matching/filtering in the future. Thanks @tadzik :)

tadzik (approved):

PR #855, opened 2020-12-11
Hernan604/dynamic builder

hernan604:
recursively builds html from data structures


gtandersen (approved):
Very cool @hernan604!

Issue #858, opened 2020-12-21
Empty Trash does not work


davidbowdley:
The Empty Trash feature has stopped working.

castaway:
Hmm, not everywhere / for everyone it seems, mine is working, which makes it harder to debug..

castaway:
Hunting around in sentry error list, I see lots of evidence though - 504, 414 and 400 errors preventing deletes - I think we shouuld create an API call for "emptytrash" instead of passing a whole long list of email ids.

PR #859, opened 2020-12-22
fix(mailview): Delete trash more efficiently (with less errors)

castaway:
Fixes: #858

Also added: updates to local index so that the changes are seen faster.

Not working (how on does this work when we delete using the canvas pop up menu?) - removing items from the canvas, if we open the Trash folder, then hit the "empty trash" button on the folder

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -513,16 +513,24 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
     });
   }
 
-  async emptyTrash(trashFolderName: string) {
-    console.log('found trash folder with name', trashFolderName);
+  async emptyTrash(trashFolder: FolderListEntry) {
+    console.log('found trash folder with name', trashFolder.folderName);
     const messageLists = await this.rmmapi.listAllMessages(
       0, 0, 0,
       RunboxWebmailAPI.LIST_ALL_MESSAGES_CHUNK_SIZE,
-      true, trashFolderName
+      true, trashFolder.folderName
     ).toPromise();
-    await this.rmmapi.deleteMessages(messageLists.map(msg => msg.id)).toPromise();
+
+    // remove local copies
+    this.searchService.deleteMessages(messageLists.map((msg) => msg.id));
+    this.searchService.updateIndexWithNewChanges();
+    // empty remote folder
+    await this.rmmapi.emptyFolder(trashFolder.folderId).toPromise();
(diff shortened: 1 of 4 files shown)


src/app/app.component.ts
```diff
@@ -513,16 +513,22 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
...
-      true, trashFolderName
+      true, trashFolder.folderName
     ).toPromise();
-    await this.rmmapi.deleteMessages(messageLists.map(msg => msg.id)).toPromise();
+
+    // remove local copies
+    this.searchService.deleteMessages(messageLists.map((msg) => msg.id));
+    this.searchService.updateIndexWithNewChanges();
+    // empty remote folder
+    await this.rmmapi.emptyFolder(trashFolder.folderId);
```
tadzik, line 528: this is missing a `toPromise()` to be effective (`emptyFolder()` returns an Observable).

tadzik (commented):
Here's a bunch of fixes I put together for it: `this.messagelistservice.setCurrentFolder()` seems like the easiest way to make it reload trash contents, which will then show up empty. It gets harder if you're just deleting individual messages from trash though – in that case I think we'll have to craft a bunch of changes and feed them to messagelistservice's `applyChanges()`.

There's another bug I've noticed though (which you may or may not want to fix here): moving messages to trash doesn't actually change the folder counter on the Trash (emptying the trash works fine, thanks to the `refreshFolderList()` you've included.

```
diff --git a/src/app/app.component.ts b/src/app/app.component.ts
index 7ec9a36..e277ca2 100644
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -525,8 +525,9 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
     this.searchService.deleteMessages(messageLists.map((msg) => msg.id));
     this.searchService.updateIndexWithNewChanges();
     // empty remote folder
-    await this.rmmapi.emptyFolder(trashFolder.folderId);
+    await this.rmmapi.emptyFolder(trashFolder.folderId).toPromise();
     // update local copies
+    this.messagelistservice.setCurrentFolder(trashFolder.folderName);
     this.messagelistservice.refreshFolderList();
     console.log('Deleted from', trashFolder.folderName);
   }
```

src/app/app.component.ts
castaway: Added, thanks!

tadzik (approved):

PR #860, opened 2020-12-22
fix(app): Hide the overview button if no local index is available

tadzik:
It doesn't work with no xapian index available, and it doesn't communicate that fact very well (at all).


castaway (approved):
Good idea!

gtandersen (approved):

PR #862, opened 2020-12-23
WIP: fix(mailviewer): Update folderlist quicker when deleting messages

castaway:
Discovered while testing "emptyTrash" fix: When deleting messages from any other folder, messages are moved to trash on the backend - the frontend just deletes them from the searchindex. It would be saner if, when deleting from anything but the trash, we instead did a searchindex move (just like movetofolder).

This doesn't work yet as I need a way to get the trash folder id..


gtandersen:
We have considered that Trash currently isn't part of the index by default (similar to Spam and Drafts)?

PR #863, opened 2020-12-23
Decouple canvastable

tadzik:
This attempts to loosen the coupling between AppComponent and CanvasTable, and hopefully eventually allow us to allow other canvastable implementations, helping with accessibility.


gtandersen:
Good initiative @tadzik!

src/app/messagelist/nativemessagelist.component.html
```diff
@@ -0,0 +1,53 @@
...
+    </div>
+    <button mat-stroked-button class="pagingButton" *ngIf="offset > 0" (click)="scrollUp()">
+        Show previous {{ rowsDisplayed }} messages
+    </button>
+    <button mat-stroked-button
+        *ngFor="let row of shownRows; index as i"
+        [ngClass]="{ 'messageButton': true, 'selectedFolder': rows.openedRowIndex == row.rowid }"
+        (click)="openMessage(row.rowid)"
+    >
+        <div class="firstRow">
```
castaway, line 13: Looks like this hard codes the set of columns displayed - which won't quite DWIM when showing search results, or the "Sent" folder.. could it use the canvastablecolumns stuff somehow?

castaway:
Any particular reason the new view isnt an HTML table? its a little confusing as a list, especially when clicking "threaded" .. cant tell what that does at all - view just jumps/redraws

castaway:
Couldnt see the "Next N messages" button at first - it disappears when the horizontal view pane is displayed

castaway:
Up/down pages the messages (by 10) on the non-canvas list.. this is quite confusing/different to the canvastable..

hernan604 (approved):

src/app/messagelist/nativemessagelist.component.html
tadzik: It does, though it cheats a little by hardcoding the header column to be either From or Sent. Not very future-proof, but that's the best we've got right now I think.

castaway:
image

Changing folders does not set the current page, so if I vist Sent, page forward a few pages, then visit Templates (or something with less pages), I get this view

castaway:
I could be missing something, or you lost it in a shuffle - when an email is open, there doesn't seem to be an indication of which one it is, in the mail list - row hightlight or similar

image

castaway:
The Attachment / Flag columns are missing - I would think they'd be possible to do?

gtandersen:
Have a look at https://material.angular.io/cdk/scrolling/overview to see if there are new strategies that are useful since this issue was last worked on.

gtandersen:
From @castaway:

1. selected some messages in the Inbox list
2. marked them as Spam
3. checked that the backend confirmed that they should now be in the Spam folder
4. some of them disappeared from the Inbox message list and .. some did not
5. also it only fetched the contents of the message after I clicked on it to view it (which was from the inbox, after it should have been moved to Spam)
image
image
image
image

PR #865, opened 2020-12-28
ci(github): Switch our CI to Github Actions

tadzik:
Seems to be a viable alternative to Travis for us, given how Travis no longer wants to do stuff for free even for opensource projects like ours.

The only downside I've found so far seems to be an inability to re-run a single job out of the batch, like we often need with our flapping e2e tests. [It's a common feature request](https://github.community/t/ability-to-rerun-just-a-single-job-in-a-workflow/17234), but no timeline yet apparently. We may want to consider expanding the e2e test runner to make it retry failed jobs, to minimize the impact that the flappiness has on our workflow.


kvviecien (approved):
Bye Travis!

Issue #867, opened 2020-12-31
Sometimes difficult to select message in list view


davidbowdley:
To select an email in list view it sometimes requires that you have to click the check box a few times before it "takes." It might be that the area you need to click on is quite small?

gtandersen:
This could be related to #694.

PR #868, opened 2020-12-31
fix(canvastable): Be a bit more lenient when detecting message selection

tadzik:
Previously it'd require you to not do *any* mouse movements between
mousedown and mouseup (while selecting a single row), which would make
it appear inconsistent. This fixes is so that the selection works even
with a slight swipe, as long as the row and column is actually the same
before and after the drag.

Fixes https://github.com/runbox/runbox7/issues/867


gtandersen (approved):
Good fix @tadzik!

castaway (approved):
Looks good to me!

PR #869, opened 2020-12-31
Account security responsiveness

tadzik:
Attempts to improve Account Security on mobile devices, aiming at little to no changes to the desktop layout.


tadzik:
Fixed toggles: 
accsec-toggles

tadzik:
Fixed TOTP instructions: 
accsec-totp

gtandersen (approved):
Great work @tadzik -- looks excellent!

tadzik:
Trusted browsers now use a responsive material table, switching to a multi-row layout on mobile screens:

trusted-desktop

trusted-mobile

tadzik:
Services now got a similar treatment:

services-desktop
services-mobile

tadzik:
Fixed app passwords:

apppass-mobile

It occured to me that a Delete button right next to a Status toggle may not be the brightest design, so I'll make all the buttons wrap to the next line to maximize the distance between them.

tadzik:
It occured to me that a Delete button right next to a Status toggle may not be the brightest design, as it'd be very easy to accidentally delete something permanently when we just want to disable it temporarily.

At first I thought it'd be a good enough idea to just make them go to their own line, but that has the effect of them being close to a thing that may want to be copied, so it becomes even worse:

delete-buttons

An alternative is to have them go their own line *and* stick to the other side:

delete-buttons-2

...and that just makes them close to the status toggle of the next item on the list, taking us back to square one (we could just slap more spacing here, but I feel like we're running in circles and not getting closer to a proper solution).

Any ideas? @gtandersen @kvviecien @davidbowdley

tadzik:
Merged into https://github.com/runbox/runbox7/pull/808.

PR #874, opened 2021-01-18
fix(compose): always ser a reply to when the "From" changes

hernan604:
Fixes a problem where the wrong reply to is sent to the server.

Situation 1 -- when it leaks the main identity email: in RMM7, if click "compose", then fill in the "To" field, then change the "From" field to my identity, it uses the last loaded reply_to

Situation 2 -- when it does NOT leak the main identity email: in RMM7, if click "compose", then change the "From" field to my identity, then fill in the "To" field, it uses the expected reply to


tadzik (approved):
In other words: when changing the From address, reply_to may not be reset correctly if the first one had it and the second one didn't?

Looks sensible to me.

PR #881, opened 2021-01-28
fix(identitys): fix error that complains from.reply_to is undefiend

hernan604:


tadzik:
I don't think this is fixing the problem: The problem is that `from` itself is undefined, and that's a symptom of a broken account, or missing data validation either around `this.draftDeskservice.froms` or `this.model.from` in there. This appears to just try to fix the symptom (and it looks bugged, since the `?:` check doesn't do anything in practice).

I think the best we can do to find the problem is to detect the situation where `from` is undefined in this piece (which is the problem we're seeing in Sentry) and then add some extra debugging to the error being thrown (since that's pretty much as much as we can do at this point).

castaway:
I think we want something along the lines of:
```
if (from && from.reply_to !== null && from.reply_to.length > 0) {
   this.model.reply_to = from.reply_to;
} 
```

We did lately change things so that reply-to doesn't get set on signup here: https://gitlab.runbox.com/runbox/rmm/-/merge_requests/398/diffs - so we should test for & expect it to be empty.

gtandersen:
Thanks @castaway -- can you have another look @hernan604? :)

Issue #886, opened 2021-02-10
Rename Webmail to Mail


gtandersen:
The main menu item Webmail can be renamed to Mail, which is currently a more common term currently. 

Any references to this item must then be updated accordingly (including documentation).

96RadhikaJadhav:
@gtandersen Can I work on it?

castaway:
Submit a pull request if you'd like to work on it

gtandersen:
This is implemented.

PR #888, opened 2021-02-11
Changed webmail to mail

96RadhikaJadhav:
This PR will resolve issue #886 .

### DIFF
--- a/src/app/app-settings.ts
+++ b/src/app/app-settings.ts
@@ -54,7 +54,7 @@ export class AppSettingsService {
     constructor(
         private storage: StorageService,
     ) {
-        this.storage.getSubject('webmailSettings').pipe(filter(s => s)).subscribe(
+        this.storage.getSubject('mailSettings').pipe(filter(s => s)).subscribe(
             (settings: any) => this.settingsSubject.next(
                 this.settings = AppSettings.load(settings)
             )
@@ -62,6 +62,6 @@ export class AppSettingsService {
     }
 
     public store(): void {
-        this.storage.set('webmailSettings', this.settings);
+        this.storage.set('mailSettings', this.settings);
     }
 }
--- a/README.md
+++ b/README.md
@@ -2,12 +2,12 @@
 
 <h4 align="center">
   <br><a href="https://runbox.com/app"><img src="src/assets/runbox7_blue_dark.png" alt="Runbox 7" width="400"></a>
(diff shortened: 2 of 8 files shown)


src/app/app-settings.ts
```diff
@@ -54,14 +54,14 @@ export class AppSettingsService {
     constructor(
         private storage: StorageService,
     ) {
-        this.storage.getSubject('webmailSettings').pipe(filter(s => s)).subscribe(
+        this.storage.getSubject('mailSettings').pipe(filter(s => s)).subscribe(
```
tadzik, line 57: This will effectively reset settings for existing users, who currently have them under `webmailSettings`. We could migrate them, but I don't think it's worth all the effort in the APIs and internals – I think the original intention was to just change it in the user-facing bits. @gtandersen?

tadzik (changes requested):

src/app/app-settings.ts
gtandersen: Yes, "any references" in #886 was meant to include items visible in user interfaces only. :)

96RadhikaJadhav: @tadzik @gtandersen I didn't get what this actually mean? Could you please let me know where exactly i have to make correction in PR?

gtandersen: Just skip the above change -- only human-readable instances of "webmail" should be changed to "mail".

96RadhikaJadhav: @gtandersen Done.

tadzik: This still breaks the settings, for example in https://github.com/runbox/runbox7/pull/888/files#diff-97880cd37962b5628664ad1984e174376d6cda35dfefb3dc0376c204516e0998R65.

The change should only concern the words that the user can see – to put it simply, those appear in the HTML files (and instances like https://github.com/runbox/runbox7/pull/888/files#diff-884f7f49640e5923f6bcac4c51d90340330a178f662defbe61e5f5aac1c512deR1130) :) In the internals it's not a problem – it's in fact useful to distinguish between a singular mail (as in, a message), and webmail the application component.

In https://github.com/runbox/runbox7/pull/888/files#diff-ce434ef5cc161194b450135a3586eb89357fc270e8aaa8e336db5a04e8d52fbeR15 both words would actually be helpful – hinting the search engines that we deal with webmail is still useful as this is a term people may use to find us.

PR #898, opened 2021-02-17
feat(payments): disable Bitpay payment option

tadzik:
It's currently not working anyway due to changes on bitpay's side.
Hopefully we can replace it with something more reliable soon.


castaway (approved):
Looks good to me

PR #901, opened 2021-02-17
fix(payments): Switch cryptocoin payments to a new, non-bitpay API

tadzik:
Could use some API renames now that it's no longer bitpay-related.


tadzik:
Also: how do we deal with emptying the cart on payments for which we won't see a confirmation any time soon?

gtandersen:
Is it critical to empty the cart if the payment is unsuccessful?

castaway (approved):
Looks fairly sane to me

PR #902, opened 2021-02-17
Last logins and sessions mobile fixes

kvviecien:
<kbd></kbd>

### DIFF
--- a/src/app/account-app/account-upgrades.component.html
+++ b/src/app/account-app/account-upgrades.component.html
@@ -51,7 +51,7 @@ mat-icon.limitedTimeOffer {
 
 <div *ngIf="rmmapi.me | async as me; else loading">
     <p>
-        <strong> Acccount: {{ me.first_name }} {{ me.last_name }} ({{ me.user_address }}) </strong>
+      <strong> Account: {{ me.first_name }} {{ me.last_name }} ({{ me.user_address }}) </strong>
     </p>
 
     <p>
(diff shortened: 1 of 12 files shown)


src/app/account-app/account-upgrades.component.html
```diff
@@ -62,55 +68,50 @@ <h2> Main account subscriptions </h2>
...
     </div>
 
     <div id="productGrid" *ngIf="subs_special | async as subs_special; else productsLoading">
-      <app-account-product
-        *ngFor="let p of subs_special"
-        [ngClass]="'productCardSpecial'"
-        [p]="p"
-        [currency]="me.currency"
-        >
+      <app-account-product *ngFor="let p of subs_special" [ngClass]="'productCardSpecial'" [p]="p"
```
tadzik, line 92: I'm not sure if this is an improvement, but alright :P

What formatter did you use?

kvviecien: Maybe i should make separate PRs for this formatting stuff. 
All I did is wrap all the pages in <section class="mat-typography"> which changes the font styles and formatting to more minimal.

kvviecien: Then I also had to change some of the headings from `h4` to `h1` to make it look the same across all the pages. 
Adding the wrapper automatically indented all the rest of the code though so that's a boomer. Wanted to avoid it but the files would look ugly.

tadzik: Rightright. No harm done :)

PR #916, opened 2021-02-22
fix(mailview): Don't display select-all checkbox on top of mail pane

### LINKED ISSUE #909: Checkbox for select all does not disappear when messages are in full height view
The checkbox for selecting all messages does not disappear when a message is opened full height and overlaps the subject area.

castaway:
No idea why I set a z-index initially!

Fixes #909


tadzik (approved):
Nice, I noticed that one just yesterday too :P

PR #919, opened 2021-02-25
feat(app): Cache message contents in indexedDB

tadzik:


src/app/xapian/searchservice.ts
```diff
@@ -528,8 +528,6 @@ export class SearchService {
             progressSnackBar.instance.messagetextsubject.next(getProgressSnackBarMessageText());
           }
 
-          this.rmmapi.deleteFromMessageContentsCache(this.pendingMessagesToProcess[this.processMessageIndex].id);
```
tadzik, line 531: The lack of this seems to break us, since changing message flags no longer invalidates the cache.

src/app/rmmapi/rbwebmail.ts
```diff
@@ -207,40 +209,49 @@ export class RunboxWebmailAPI {
...
+            // This kind of multilevel caching may seem excessive,
+            // but it turns out that pulling a single row from indexedDB (or maybe from Dexie specifically?)
+            // can easily take over 100ms: the difference between "instant" and "fast"
+            // Since instant is the only acceptable result, we cache indexedDB lookups here.
 
-            return messageContentsObservable;
+            cached = this.messageContentsRequestCache[messageId] = this.messageCache.get(messageId);
         }
+        return cached;
     }
```
castaway, line 230: It doesn't seem to limit the messageContentsRequestCache in any way - it would probably be sane to have a max number of items in it (tracked by last used date?) .. just in case people scroll through pages of emails then complain that their browser explodes

tadzik: Add an LRU in 90625739e0c0fa45803f0358c9d5e91f087a6e93.
We could probably make it a little more generic for more varied uses if we decide that we need that sort of thing in other components. Currently it'll hold a maximum of 100 messages at a time (the rest will remain in IndexedDB), and enforce the cache limit no sooner than 1 second after the last insertion.

castaway:
If I load the current state, watch the "Network" tab in devtools, and scroll down a long folder, then back up - I'd expect it not to refetch a bunch of messages that it should have fetched on the way down .. But it does?

Also seems to request some several times in a row:

castaway:
Retested - before we were downloading content for each message visible in the message list, as I scrolled - now we're not doing that at all, it doesn't fetch the message contents until I click on them. Did the sorting of the issue above go to far, or is this the new plan?

Determined by scrolling the message list, and watching the network activity for /rest/v1/email/3909775 style calls.

castaway:
Retested by loading page, then deleting indexdb for messageCache - this makes maaany errors (is fine if I reload the page again..) - mebbe should cope better with that? Otherwise does seem to fetch (and not refetch) as expected now

Issue #713, opened 2020-08-06
Deleting multiple messages resulted in an empty message list


runboxdave:
I selected and deleted a number of messages and after the delete option the screen that was returned was empty of messages. I waited but nothing happened. However, when I touched the screen the message list appeared. It would seem that the screen just needed to refresh.

gtandersen:
Confirmed that this happens intermittently, but no reliable pattern detected yet.

castaway:
The "deleteMessages" action doesn't seem to do anything that triggers an immediate screen update, so it would end up waiting on the next index refresh, I think. (which could be 10secs away)

gtandersen:
It does trigger an immediate update when the local index is in use, but not otherwise.

gtandersen:
@davidbowdley Did this occur while using the local index?

I have reproduced a similar symptom where the message list disappears beneath a message that is deleted (with local index), but it doesn't happen consistently. I then have to click to another folder and then back again for the entire list to be displayed.

On closer inspection it appears that when you delete a message, a blank full height box is inserted in place of the deleted entry and the vertical scrollbar disappears. When you scroll down, the next messages appear beneath this blank box.

gtandersen:
Another symptom that appears to correlate is that the padlock symbol in the URL bar is broken (seemingly insecure elements). A forced reload of the app appears to resolve both symptoms until it occurs again.

gtandersen:
Further information: It appears the problem might be redrawing the entries beneath the deleted ones, as those entries sometimes appear after a short delay.

gtandersen:
Check for console errors when this happens.

gtandersen:
Probably related to one of these commits: 

https://github.com/runbox/runbox7/commits/master/src/app/canvastable/canvastable.ts

gtandersen:
This continues to happen and there are no errors in the console. Just seeing the usual `Getting latest messages from server` and other typical output.

gtandersen:
This continues to happen intermittently, i.e. deleting or moving messages doesn't successfully update the message list and the message list is stuck either with no change at all or a partially empty message list.

gtandersen:
It does not appear related to performance on the local device, but possibly response from REST servers.

gtandersen:
See also internal issue 1156.

gtandersen:
This is also reported in smaller accounts, ref. support ticket DUY-755-41975.

castaway:
Does this happen in/from specific folders? With search index on or off?

[...]

PR #920, opened 2021-03-02
fix(mailview): Ensure we update everything when deleting messages

castaway:
(Maybe) fixes: #713

So this went a bit crazy: This patch tweaks most of the user ui -> API interaction regarding mail, statuses, maillist display etc, so that we :

1. First update the ui - eg search index, message list, folder counts etc, locally
2. Then post the change to the backend, and call update Index to merge in anything else found via that update

This means the UI updates pretty fast to indicate what the user has done, but if the API call has an error (and it seems some report custom errors, some don't!), the UI may seem to change, and then change back as we fetch the current index / folder state and merge back in changes that were supposed to happen but didn't. (NB: I havent encountered this in my testing, APIs failing, so this is hopefully mostly a cautionary tale)


src/app/mailviewer/rmm7messageactions.ts
```diff
@@ -35,6 +36,15 @@ export class RMM7MessageActions implements MessageActions {
     draftDeskService: DraftDeskService;
     rmmapi: RunboxWebmailAPI;
 
+    public updateMessages(messageIds: number[],
+                          updateIndex: (messageIds: number[]) => void,
+                          updateAPI: (messageIds: number[]) => Observable<any>,
+                         ): Observable<any> {
+        updateIndex(messageIds);
+        updateAPI(messageIds).subscribe(() => this.searchService.updateIndexWithNewChanges());
+        return of(null);
```
tadzik, line 45: Any reason why it returns an observable? It's not dependent on the result of previous, async actions, so in practice the function is as good as if it was a synchronous one, I think. If we wanted it to be a "notify me when finished" thing, we'd have to pipe off of updateAPI or so.

I like the overall idea, but I'd reconsider Observables here – Promises might be more fitting, since they actually execute their actions even if you don't subscribe to them, and also don't have a habit of executing twice if you subscribe to them twice :P

castaway: The idea was that after all this was actually finished, the calling func can do things like "remove selections" etc

castaway: .. right, realised it didnt actually need to do that at all (at least not in this case) .. updated! now to review the others..

src/app/app.component.ts
```diff
@@ -599,69 +638,81 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
...
+      // remote api
+      (msgIds: number[]) => {
+        const req = this.rmm.email.update({
+          flag: {
+            name: 'seen_flag',
+            value: status ? 1 : 0,
+          },
+          ids: msgIds
+        });
+        req.subscribe(() => snackBarRef.dismiss());
```
tadzik, line 662: This may be a good fit for a `.pipe(tap(() => snackBarRef.dismiss()))` instead of a `subscribe()`, which would ensure that it doesn't get executed twice (even if it weren't `shared()`).

src/app/app.component.ts
```diff
@@ -599,69 +638,81 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
...
+          this.rmmapi.messageFlagChangeSubject.next(
+            new MessageFlagChange(id, status, null)
+          );
+        });
         this.clearSelection();
-        this.showSelectMarkOpMenu = false;
-        snackBarRef.dismiss();
-    });
+      },
+      // remote api
```
tadzik, line 653: Comments similar to these pop up in most of these `updateMessages()` calls – they are helpful, which makes me wonder if we want to enforce this in an API using quasi-named arguments, like so?

```
this.messageActionsHandler.updateMessages(messageIds, {
    local: fun1(),
    remote: fun2()
});
```

What do you think?

tadzik: Given our usage patterns here, it could also have a `finally` optional callback, that'd prevent us from subscribing/piping the result to react to remote changes.

src/app/app.component.ts
```diff
@@ -854,36 +907,50 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
 
   dropToFolder(folderId): void {
     const messageIds = this.canvastable.rows.selectedMessageIds();
+    let folderPath;
+    this.messagelistservice.folderListSubject.subscribe(folders => {
+      folderPath = folders.find(fld => fld.folderId === folderId).folderPath;
```
tadzik, line 912: I don't think we can count on this actually being executed synchronously – to be safe, everything that uses `folderPath` below should be a part of this `subscribe()` callback.

src/app/app.component.ts
```diff
@@ -854,36 +907,50 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
...
       if (folder) {
-        this.searchService.updateIndexWithNewChanges();
-        this.clearSelection();
-        this.canvastable.rows.clearOpenedRow();
+        let folderPath;
+        this.messagelistservice.folderListSubject.subscribe(folders => {
+          folderPath = folders.find(fld => fld.folderId === folder).folderPath;
+        });
+
+        console.log('Moving to folder', folderPath, messageIds);
```
tadzik, line 940: Same as above: this could be uninitialized.

src/app/mailviewer/rmm7messageactions.ts
```diff
@@ -17,32 +17,60 @@
...
 
-        dialogRef.componentInstance.selectedMessageIds = [this.mailViewerComponent.messageId];
         dialogRef.afterClosed().subscribe(folder => {
             if (folder) {
-                this.searchService.updateIndexWithNewChanges();
-                this.mailViewerComponent.close();
+                let folderPath;
+                this.messageListService.folderListSubject.subscribe(folders => {
+                    folderPath = folders.find(fld => fld.folderId === folder).folderPath;
+                });
```
tadzik, line 59: Similar uninitialized folderPath worry as above :)

src/app/rmmapi/messagelist.service.ts
```diff
@@ -229,6 +242,62 @@ export class MessageListService {
         }
     }
 
+    // When emptying the trash we delete from here first
+    // (then update backend + index)
+    public pretendEmptyTrash() {
```
tadzik, line 247: Clever :)

src/app/rmmapi/rbwebmail.ts
```diff
@@ -412,18 +412,21 @@ export class RunboxWebmailAPI {
...
-                new MessageFlagChange(messageId, seen_flag_value === 1 ? true : false, null)
-            )),
-            mergeMap(() => this.http.put('/rest/v1/email/' + messageId, JSON.stringify({ seen_flag: seen_flag_value }))
-        ));
+        // return of(null).pipe(
+        //     tap(() => this.messageFlagChangeSubject.next(
+        //         new MessageFlagChange(messageId, seen_flag_value === 1 ? true : false, null)
+        //     )),
+        //   mergeMap(() =>
+            return this.http.put('/rest/v1/email/' + messageId, JSON.stringify({ seen_flag: seen_flag_value }));
```
tadzik, line 428: Note that this HTTP query will not actually execute unless the caller subscribes to it.

src/app/xapian/searchservice.ts
```diff
@@ -602,6 +602,9 @@ export class SearchService {
    * @param destinationfolderPath
    */
   moveMessagesToFolder(messageIds: number[], destinationfolderPath: string) {
+    if (!this.api) {
```
tadzik, line 605: Does it deserve a warning here maybe? We could also wait for `initSubject()` to make sure we execute it as soon as it's possible.

tadzik (commented):
Spotted a few potential bugs, so leaving an RC :) I like what I'm seeing though, much saner and more consistent, even in the code alone.

src/app/xapian/searchservice.ts
castaway: This is the "if not using local index" guard, as used elsewhere in the code (it was just missing here for some reason)

src/app/rmmapi/rbwebmail.ts
castaway: Yup, like most of the message action api calls

src/app/mailviewer/rmm7messageactions.ts
```diff
@@ -17,43 +17,76 @@
...
 export class RMM7MessageActions implements MessageActions {
     mailViewerComponent: SingleMailViewerComponent;
     dialog: MatDialog;
     snackBar: MatSnackBar;
     searchService: SearchService;
+    messageListService: MessageListService;
     draftDeskService: DraftDeskService;
     rmmapi: RunboxWebmailAPI;
 
+    public updateMessages(args: { messageIds: number[],
```
tadzik, line 41: This type definition has become a bit bulky, maybe we can just name a type after it? :)

tadzik (approved):
Some leftover comments I'd drop, otherwise looks good :)

PR #921, opened 2021-03-03
feat(api): api docs component

hernan604:


castaway (approved):
Looks good - needs de-conflicting

PR #923, opened 2021-03-08
Fix build errors

tadzik:
And some merge screwups


castaway (approved):
looks good to me (feeling a bit guilty over here)

PR #927, opened 2021-03-12
feat(account): Split renewals page into active and expired products

tadzik:
This should make it easier to tell the ones the user cares about from the ones that are long gone, especially in older accounts.

Looks like this:

active

expired

The order of importance is: the soonest-expiring first on the Active tab, latest-expired first on the Expired tab.

I do have a problem with the naming though. The Active/Expired naming makes sense to me, but we call the entire page Active products, which now makes it awkward. This was a problem with variable naming in the source as well, which I noted down in the comment. @davidbowdley suggested renaming the entire page to Products – on the other hand, the whole Account section is called "Account products", referring to both available and, ugh, active-and-expired ones. Ideas?

### DIFF
--- a/src/app/account-app/account-renewals.component.html
+++ b/src/app/account-app/account-renewals.component.html
@@ -6,149 +6,145 @@
         </app-runbox-loading>
     </div>
 
-    <ng-template #productsTable>
-
-        <p>
-            Here you can see all the active products in your account, and renew any of them at any time.
-        </p>
-
-        <table mat-table [dataSource]="active_products" [ngStyle]="{ width: '100%' }" multiTemplateDataRows>
-            <ng-container matColumnDef="name">
-                <th mat-header-cell *matHeaderCellDef> Name </th>
-                <td mat-cell *matCellDef="let p">
-                    {{ p.name }}
-                    <strong *ngIf="p.pid === current_subscription"> – your current subscription </strong>
-                </td>
-            </ng-container>
-            <ng-container matColumnDef="quantity">
-                <th mat-header-cell *matHeaderCellDef> Quantity </th>
-                <td mat-cell *matCellDef="let p">
-                    {{ p.quantity }}
-                    <span *ngIf="p.subtype === 'subaccount'">
-                        ({{ p.subaccounts.length }} in use)
-                        <button mat-icon-button *ngIf="p.subaccounts.length > 0" (click)="showSubsDialog(p)">
-                            <mat-icon svgIcon="information-outline"
-                                [matTooltip]="'Associated accounts: ' + p.subaccounts.join(', ')"></mat-icon>
-                        </button>
-                    </span>
-                </td>
-            </ng-container>
-            <ng-container matColumnDef="active_from">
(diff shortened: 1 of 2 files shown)


gtandersen:
Nice work @tadzik!

One solution (as discussed elsewhere) could be to rename the section to Subscriptions and remove the concept of products altogether.

castaway:
Not sure you changed this one, but the titles/menu entries for "Main Account Subscriptions/Plans" are confusing:
image

davidbowdley:
It's quite important that recently expired products should still appear under "Active". Perhaps for 1 month? Then they can move to Expired. They shouldn't appear in both places during this time.

gtandersen:
Agree that we could be even more consistent wrt naming here.

See also https://github.com/runbox/runbox7/issues/424#issuecomment-577504496 for a discussion about naming (where Products can be replaced with Subscriptions).

Note that aliases need to remain a product/quota at least for Runbox addresses since aliases reserve addresses that are in demand.

I think Transactions can remain its own menu item and page since it's an important record a customer might expect to find easily.

davidbowdley:
We should look at the URLs too as they may need changing. You go up /upgrades to downgrade/change your plan and /renewals to see your active products while /credit-cards may also be debit cards (which are by far more common).

Sub-accounts are probably the most commonly purchased add-on so should be listed first on the add-ons page.

There are too many clicks required (four I believe from the mail webmail screen) before you see a product to purchase.

gtandersen:
I like that each page isn't too long as it makes them easier to navigate, but perhaps we could merge the two pages Main Account Plans and Add-ons & Sub-Accounts since they are closely related and the product cards could possibly be made smaller (and fewer), thusly:

1. Main account subscriptions 
2. Sub-account subscriptions
3. Main account add-ons (these can probably be phased out over some time)

And I agree that it would be sensible to show existing products/subscriptions first in the menu, but if we want separate pages for Transactions and Card Details these would logically come next and put Main Account Plans and Add-ons & Sub-Accounts last.

So, how about the following menu (with the existing titles in parenthesis and proposed URL thereafter):

**Subscriptions & Payments**
- **Plans & Upgrades** (Main Account Plans + Add-ons & Sub-Accounts): /app/account/plans
- **Your Subscriptions** (Active products & Renewals): /app/account/subscriptions
- **Your Payments** (Transactions): /app/account/payments
- **Your Cards** (Card Details): /app/account/cards

Alternatively, if we want to list existing subscriptions first, we could use this menu:

**Subscriptions & Payments**
- **Your Subscriptions** (Active products & Renewals): /app/account/subscriptions
- **Plans & Upgrades** (Main Account Plans + Add-ons & Sub-Accounts): /app/account/plans
- **Payment History**  (Transactions): /app/account/payments
- **Payment Cards** (Card Details): /app/account/cards

By the way, I think we decided to move away from tabs due to issues on smaller screens, so Active Products and Expired products could perhaps be vertical sections instead? :)

gtandersen:
We decided to go with the second option for **Subscriptions & Payments** above.

PR #929, opened 2021-03-15
style(payment): Replace cryptocurrency logo and improve formatting.

gtandersen:


gtandersen:
Payment methods

tadzik:
Looks good, and scales well to mobile screens too :)

I wonder if we want to mention Coinbase on the main card though: it's basically an implementation detail, the customer just does a cryptocoin transfer like any other and they never need to know Coinbase exists, unless they explicitely choose to use their webui in the actual payment dialog:

coinbase

I bring this up because Bitpay previously confused people to the point where they wondered if they need a Bitpay account, a Bitpay app or something along those lines, if I remember correctly (@davidbowdley can probably shed some light). Hence, I think we may want to "hide" that fact a little to not imply that this is something else than a standard cryptocurrency payment.

gtandersen:
Right, I included "Coinbase" on the payment method card to be consistent with the use of "Stripe" for card payments, and although Coinbase is somewhat irrelevant in this context both names might serve a purpose in providing a sense of security (since they are at the present probably more known than our own).

Another option could be to replace the text with something like "Pay with cryptocoins such as BTC and ETH", unless it's too long for one line of text.

tadzik:
Makes sense. Approving this now, and leaving up for a bit for others to possibly comment :)

tadzik (approved):

tadzik:
Tests are all green locally.

PR #931, opened 2021-03-16
feat(app): Use a new API to pre-cache messages in bulk

tadzik:
Needs to be similarly adopted in the SearchService.


src/app/rmmapi/rbwebmail.ts
```diff
@@ -222,14 +222,62 @@ export class RunboxWebmailAPI {
...
+                (err: Error) => {
+                    for (const msgid of missingMessages) {
+                        this.messageContentsCache[msgid].error(err.toString());
+                        delete this.messageContentsCache[msgid];
+                    }
+                }
+            );
+        }
+
+        //return Promise.allSettled(messagePromises);
```
tadzik, line 277: Not sure yet if we'd actually prefer `allSettled()` here or not, but it doesn't compile under our current setup so far anyway ;) If we go for it, we might need a polyfill to make it work on ancient browsers too.

gtandersen (approved):
Looks good @tadzik!

castaway (commented):
Requires #972 to be updated/deployed first

PR #936, opened 2021-03-19
fix(account): Make sure subaccounts can access account security and other not payment related paths

tadzik:


gtandersen (approved):
Looks good @tadzik!

castaway (approved):
LGTM

PR #937, opened 2021-03-21
Condensed Overview layout with new selector for sorting.

gtandersen:
Needs to be wired up. :)


tadzik:
Wired up here: https://github.com/gtandersen/runbox7/pull/4 :)

kvviecien (approved):
Seems like nothing changed here since the last time I looked, so everything's good to go!

PR #940, opened 2021-03-24
feat(settings): Add content to the main Settings page

kvviecien:
<kbd>localhost_4200_account
</kbd>


tadzik:
Looks great (on mobile too), though the font seems a bit small? Maybe it's just me.

One thing that stands out is that it's currently unreachable except for mobile­ – in the desktop view there is no link to this page in the entire Settings dropdown (that I can find). Perhaps we should make the desktop menu behave the same way, with no dropdown at all, consistently with the rest of the menubar? The /account/components page can probably be retired then.

kvviecien:
> Looks great (on mobile too), though the font seems a bit small? Maybe it's just me.
> 
> One thing that stands out is that it's currently unreachable except for mobile­ – in the desktop view there is no link to this page in the entire Settings dropdown (that I can find). Perhaps we should make the desktop menu behave the same way, with no dropdown at all, consistently with the rest of the menubar? The /account/components page can probably be retired then.

I feel like the screenshot made it look weird. The font is exactly the same as in all the paragraphs in the rest of Settings.

I actually have the changes ready to remove the drop-down menu on another branch. I will push it to this one then, so we can have everything ready. And will delete the components page too :)

tadzik (approved):

PR #945, opened 2021-03-31
Account details pages

kvviecien:
Upgrade Account Details page to Runbox7, which is split into 3 pages now:

- Account Details 
- Account Settings
- Storage and Data

_DRAFT. Needs API wiring and some security work._

### DIFF
--- a/src/app/account-app/account-app.component.html
+++ b/src/app/account-app/account-app.component.html
@@ -109,13 +109,21 @@
                             <mat-icon mat-list-icon svgIcon="account"></mat-icon> Account & Security
                         </p>
                     </mat-expansion-panel-header>
-                    <a href="/mail/account" target="rmm6">
-                        <mat-list-item [matTooltip]="rmm6tooltip">
-                            <p mat-line>
-                                <mat-icon mat-list-icon svgIcon="details"></mat-icon> Account Details
-                            </p>
-                        </mat-list-item>
-                    </a>
+                    <mat-list-item routerLink="/account/details">
+                        <p mat-line>
+                            <mat-icon mat-list-icon svgIcon="face"></mat-icon> Personal Details
+                        </p>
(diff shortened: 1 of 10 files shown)


gtandersen:
Great work @kvviecien -- are there any screenshots available? :)

kvviecien:
> Great work @kvviecien -- are there any screenshots available? :)

@gtandersen Yes! Here you go: 

<kbd></kbd>
<kbd></kbd>
<kbd></kbd>

kvviecien:
Here I was thinking that maybe instead of having a password field on the top, we could have a pop-up like in Account Security asking to provide current password?

tadzik:
> Here I was thinking that maybe instead of having a password field on the top, we could have a pop-up like in Account Security asking to provide current password?

Sounds like a good idea, even for consistency's sake.

gtandersen:
Good idea @kvviecien!

gtandersen:
@kvviecien The screenshots look great! 

Perhaps the fields that expect less input could be shorter, such as phone number, organization number, etc? :)

kvviecien:
> @kvviecien The screenshots look great!
> 
> Perhaps the fields that expect less input could be shorter, such as phone number, organization number, etc? :)

Maybe let's shorten all of them? They're indeed a little longer than needed in most cases, but changing only few could affect the look too much. Consistency again :)

gtandersen:
> Maybe let's shorten all of them? They're indeed a little longer than needed in most cases, but changing only few could affect the look too much. Consistency again :)

Agreed!

gtandersen:
@kvviecien How are we doing with this -- are there still backend changes needed?

gtandersen:
API endpoints are in place; working to integrate npm package for countries/timezones.

kvviecien:
Waiting for hernan's changes to be deployed on Monday.

PR #947, opened 2021-04-04
style(security): Enlarge toggles and improve formatting.

gtandersen:
Runbox 7 2FA toggles

### DIFF
--- a/src/app/runbox-components/runbox-section.ts
+++ b/src/app/runbox-components/runbox-section.ts
@@ -28,7 +28,7 @@ import { RMM } from '../rmm';
     `],
     template: `
     <div class="app-runbox-section">
-        <div style="background: #e1eeff; padding: 15px;"
+        <div style="background: #e1eeff; padding: 15px 0;"
             [ngStyle]="{'background': heading_sizes[size].background}"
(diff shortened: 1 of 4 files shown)


tadzik (approved):
Looks good!

src/app/runbox-components/runbox-section.ts
```diff
@@ -47,39 +47,33 @@ export class RunboxSectionComponent {
...
     h4: {
-        background: '#7BB1FA'
     },
     h5: {
-        background: '#549CFC'
     },
     h6: {
-        background: '#207efe'
     },
   };
```
kvviecien, line 78: After removing all those background colors, the whole `heading_sizes` block isn't needed there anymore. Maybe it's best to just remove the whole thing, keeping it clean :)

kvviecien (changes requested):

src/app/runbox-components/runbox-section.ts
gtandersen: That's true, but it might be useful with the placeholders if we'd like to add styles in the future. :)

kvviecien (approved):

PR #951, opened 2021-04-12
feat(support): Added bug report button with a compose template

castaway:


tadzik (approved):
Looks good :)

I'd consider abstracting the https://github.com/runbox/runbox7/pull/951/files#diff-8d554c98f4c32fee762d6a18e8e189f8898124312ba390bdc51fadd38345674fR132 bit into its own component (`RunboxNavListItem`?) since we seem to repeat that sequence quite a bit, including in [other apps](https://github.com/runbox/runbox7/blob/master/src/app/contacts-app/contacts-app.component.html#L35)

src/app/app.component.html
```diff
@@ -129,6 +129,13 @@
       </mat-list-item>
     </mat-nav-list>
 
+    <mat-nav-list>
+      <mat-list-item (click)="composeBugReport()" matTooltip="Compose an email to runbox support">
```
tadzik, line 133: Is there any reason this is its own nav-list, rather than another list-item in the previous nav-list?

castaway: Only cos that one had an ngIf attached, I guess we could move that to the nav-list-item

gtandersen (approved):
Nice work @castaway!

PR #953, opened 2021-04-13
fix(canvastable): Make inline message previews work again

tadzik:
Here's to hoping this temporary hack doesn't stay there forever :)


castaway (approved):
mmm tasty cache layers

PR #957, opened 2021-04-15
feat(rmm): add pubsub pattern to rmm.profiles

hernan604:
can be used by the draftdesk to get signaled when identitys are loaded/reloaded. subscribe from inside the draftdesk. ie:

var subscription1 = profile.pubsub.subscribe('profiles_verified.updated', function ( profiles_verified ) {
  console.log( 'profiles_verified has changed, do something' );
});


gtandersen (approved):
Looks good @hernan604 -- can you have a look as well @tadzik?

castaway:
We solved this in #967 (already merged)

PR #960, opened 2021-04-15
fix(mailviewer): Sent folder now ignores "unread only" setting

### LINKED ISSUE #893: Unread option is also being applied to the Sent folder
The option to show only unread messages is also being applied to the Sent folder and this isn't appropriate. The result is you see no messages in your Sent folder.

castaway:
Fixes #893

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -1023,7 +1023,10 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
           default:
             if (this.searchText.length < 3) {
               // Expand to all folders if search text length is longer than 3 characters
-              querytext += this.searchService.getFolderQuery(querytext, this.selectedFolder, this.unreadMessagesOnlyCheckbox);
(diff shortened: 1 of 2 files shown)


gtandersen (approved):

src/app/xapian/searchservice.ts
```diff
@@ -662,7 +662,7 @@ export class SearchService {
       querytext += folderQuery(folderPath);
     }
 
-    if (unreadOnly) {
+    if (unreadOnly && !this.messagelistservice.ignoreUnreadInFolders.includes(folderPath)) {
```
tadzik, line 665: I'm not too fond of this level of intimacy between this method and MessageListService.

Would it be very annoying to just not set unreadOnly when we're querying for the Sent folder?

castaway: Not particularly...

castaway: Logic moved up a level

tadzik (approved):

PR #964, opened 2021-04-20
fix(mailviewer): Maintain selected messages when new mail appears

### LINKED ISSUE #914: Maintain checkbox selection on message list update
When selecting one or more checkboxes in the message list and the list is updated with new messages before further action is taken, the selected checkboxes are automatically moved up the list accordingly.

Instead, the same checkboxes should remain selected to prevent action to be taken on the wrong messages.

castaway:
Fixes #914

### DIFF
--- a/src/app/common/messagedisplay.ts
+++ b/src/app/common/messagedisplay.ts
@@ -21,8 +21,9 @@ import { CanvasTableColumn } from '../canvastable/canvastablecolumn';
 export abstract class MessageDisplay {
   public openedRowIndex: number;
   public selectedRowId: number;
-//  public openedRowId: number;
-  public selectedRowIds: { [key: number]: boolean } = {};
+  //  public openedRowId: number;
+  public msgIdsSelected: { [key: number]: boolean } = {};
+  // public selectedRowIds: { [key: number]: boolean } = {};
   public hasChanges: boolean;
 
   public rows = [];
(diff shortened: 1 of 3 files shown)


gtandersen (approved):
Excellent, @castaway!

src/app/common/messagedisplay.ts
```diff
@@ -36,20 +37,38 @@ export abstract class MessageDisplay {
...
+    return Object.keys(this.msgIdsSelected).filter((msgId) => this.msgIdsSelected[parseInt(msgId, 10)]).map((msgId) => parseInt(msgId, 10));
   }
 
   allSelected(): boolean {
-    return Object.keys(this.selectedRowIds).filter((key) => this.selectedRowIds[key]).length === this.rows.length;
+    return Object.keys(this.msgIdsSelected).filter((key) => this.msgIdsSelected[key]).length === this.rows.length;
+  }
+
+  anySelected(): boolean {
+    return Object.keys(this.msgIdsSelected).reduce((prev, current) =>
```
tadzik, line 54: I think this could be simplified to `!!Object.keys(this.msgIdsSelected).find(id => this.msgIdsSelected)` or something similar.

src/app/common/messagedisplay.ts
```diff
@@ -36,20 +37,38 @@ export abstract class MessageDisplay {
     return this.rows.length;
   }
 
+  selectedRowIds(): number[] {
+    return Object.keys(this.selectedMessageIds).map((msgIndex) => this.findRowByMessageId(parseInt(msgIndex, 10)));
+  }
+
+  // msgIds where the value is true
```
tadzik, line 44: What value? ;)

src/app/common/messagedisplay.ts
tadzik: Or even `this.selectedMessageIds().length > 0`

src/app/common/messagedisplay.ts
```diff
@@ -36,20 +37,38 @@ export abstract class MessageDisplay {
     return this.rows.length;
   }
 
+  selectedRowIds(): number[] {
+    return Object.keys(this.selectedMessageIds).map((msgIndex) => this.findRowByMessageId(parseInt(msgIndex, 10)));
```
tadzik, line 41: Isn't `this.selectedMessageIds` a method? Did this mean to use this.`msgIdsSelected`?

This could be a good argument to give these more obvious names, right now we have attributes that look very similar to methods.

src/app/common/messagedisplay.ts
castaway: Could even just be a "find where any is true" .. I just moved this code from app.component ... didnt re-read it ;)

src/app/common/messagedisplay.ts
```diff
@@ -36,20 +37,38 @@ export abstract class MessageDisplay {
...
+  }
+
+  // msgIds where the value is true
   selectedMessageIds(): number[] {
-    return Object.keys(this.selectedRowIds).map((rowIndex) => this.getRowMessageId(parseInt(rowIndex, 10)));
+    return Object.keys(this.msgIdsSelected).filter((msgId) => this.msgIdsSelected[parseInt(msgId, 10)]).map((msgId) => parseInt(msgId, 10));
```
tadzik, line 46: I don't think these parseInts are necessary: unless we lie in the type definition of the attribute and these isn't actually `{ [number]: boolean }` ;)

tadzik (changes requested):

src/app/common/messagedisplay.ts
castaway: Object.values(this.msgIdsSelected).includes(true); ? feels like the shortest one

src/app/common/messagedisplay.ts
castaway: *the* value! its a hash ;)

src/app/common/messagedisplay.ts
castaway: Yes, oops (I renamed it mid fix)

src/app/common/messagedisplay.ts
castaway: Object.keys makes things back into strings :(

tadzik (approved):

PR #995, opened 2021-05-14
Bug report fixes

gtandersen:


davidbowdley (approved):

src/app/compose/draftdesk.service.ts
```diff
@@ -261,16 +261,14 @@ export class DraftDeskService {
...
             'Runbox 7 bug report'
         );
         const template = await this.http.get('assets/templates/bug_report.txt',
                                              {responseType: 'text'}).toPromise();
         const me = await this.rmmapi.me.toPromise();
 
-        let body = `
-${template}
-`
+        let body = `${template}`
```
tadzik, line 271: This could probably just be `let body = template` at this point :)

src/assets/templates/bug_report.txt
```diff
@@ -1,48 +1,71 @@
...
+19. Mail viewer on right (auto-detected): %%SETTINGVIEWRRIGHT%%
+
+20. Show unread email only (auto-detected): %%SETTINGUNREAD%%
+
+21. Any other relevant changes you have made to Runbox 7 settings:
+
+
+You may attach screenshots to your message if you think they might help us investigate the error.
+
+Ensure that the subject line of this message is unchanged, and then click the Send button to file your bug report.
```
tadzik, line 64: We can consider locking the subject field when these are being sent (might be a bit involved).

tadzik (approved):

src/app/compose/draftdesk.service.ts
castaway: I'm sure there was a point to the newlines, tho I've now forgotten why, I think I originally planned to use variables for the values in the template and interpolate them this way too - but didn't. If newlines not required then yes, just use the template var instead ;)

castaway (approved):

src/app/compose/draftdesk.service.ts
gtandersen: The newlines added newlines in the template too, which caused the text to appear on line 2 in the message. Removing them fixed that issue while not seeming to introduce any new ones. :)

Issue #994, opened 2021-05-12
Payment Cards page additional information.


davidbowdley:
For most people the Payment Cards page will mostly be empty. We could include some of the information here:

https://runbox.com/price-plans/payment-methods/

Or link to the page (or both).

gtandersen:
@davidbowdley Can you specify the exact text you think we should include? :)

gtandersen:
This is done:

<img width="831" alt="Paymente Cards" src="https://user-images.githubusercontent.com/1620058/122653094-536bb800-d0f7-11eb-9aef-58c656e5ee3c.png">

Issue #990, opened 2021-05-12
Titles and URLs in "Subscriptions & Payments"


runboxdave:
"Your Subscription"s goes to a page called "Account products" 

"Payment Cards" goes to a page called "Your credit cards" which may not be inclusive of all kinds of payment cards (not even the majority in circulation) and so it might be better if it is also called "Payment Cards" for consistency. The URL could also be updated as it is currently /credit-cards

kvviecien:
@runboxdave apart from the link update, is there anything else that has to be done still? Looks like we covered it over the time.

runboxdave:
@kvviecien The URL for "Payment Cards" is still /credit_cards which is a minor point but more debit cards are in circulation than credit cards (by far) and so it is probably worth changing to /cards or /payment_cards to avoid any possible confusion.

The subscriptions title and URL now match :)

PR #996, opened 2021-05-14
style(payment): Improve text and icons.

### LINKED ISSUE #992: Payment History page shows a column called Time but it's actually a Date in that column
The "Payment History" page shows a column called Time but it's actually a Date in that column.

gtandersen:
Fixes #992, #994, and #990 (partially).


davidbowdley (approved):

gtandersen:
Runbox 7 Your Subscriptions
Runbox 7 Plans   Upgrades
Runbox 7 Payment History
Runbox 7 Payment Cards

gtandersen:
Runbox 7 Payment Cards

src/styles.scss
```diff
@@ -123,19 +129,43 @@ ul.square-list li {
     font-weight: 400;
 }
 
+.mat-raised-button:hover {
+    background-color: mat-color($rmm-default-highlight);
+}
```
tadzik, line 134: We should also set the foreground colour here: otherwise, (maybe only on a dark system theme?) the background becomes white and the text colour also becomes white:

hover

src/app/account-app/account-app.component.html
```diff
@@ -17,8 +31,9 @@
         [fixedInViewport]="mobileQuery.matches"
         fixedTopGap="0"
         id="sideMenu"
-        style="width: 330px"
-    >
+		 >
+        <h3>Settings Menu</h3>
```
tadzik, line 35: This looks a bit out of place imo (and we don't have a similar "menu title" on any other page).

If we want to make sure that the user knows that they're in the settings, we could try making the header toolbar Settings icon stand out somehow when it's the page we're on (like showing the text not only on hover)?

tadzik (changes requested):

src/styles.scss
gtandersen: This should now be fixed by reverting this style and instead using the `color` parameter that is used elsewhere in the app.

It's possible all raised buttons should have the primary color but there might ones that should be excluded, so I refrained from adding a global style at this point.

src/app/account-app/account-app.component.html
gtandersen: I was ambivalent about adding it, but reasoned as follows:

1. It is important that the user immediately understands that the left-pane menu is used to navigate Settings and not the rest of the app, since the items will increasingly mirror the main menu items. (The need to clearly set Settings apart from the rest of the app was one reason for suggesting an "overlay" effect when accessing Settings.)
2. The main content area needed top padding to add vertical space above the heading "Account Settings". Consequentially the left-pane menu looked oddly attached to the top header, and equally odd with a similar padding without a heading on its own.
3. None of the other areas of the app (Mail, Contacts, Calendar) have a *menu* per se; although they have left panes, these mainly contain items for navigating various data selections (aside from the individual Settings items which should in the future take the user to Account Settings).
4. Adding similar headings in the other areas of the app would certainly look misplaced, and I don't think we should do that.

In short: I agree that adding a Settings Menu heading breaks convention but that it might be warranted, unless a better solution can be found that take the above 4 concerns into account.

I also agree with the suggestion to indicate which main menu item has been selected, and presumably this requires some Angular logic in order to be robust? :)

davidbowdley: I think it's clear that you are in settings menu as you click the settings icon to get there. I'm not sure if there was a question here about whether there should be a settings icon in the main menu but if there was I think it is correct it is there as Settings is a part of the app just like Mail, Calendar etc. and it has a section of it's own just like the others.

I don't think there is much to get confused about here once you've clicked on the Settings icon once (with or without a heading) :)

gtandersen: > I think it's clear that you are in settings menu as you click the settings icon to get there. 

Not necessarily, as you could also (in the near future) have clicked a button elsewhere in the app (e.g. Mail or Contacts), or followed a link to https://runbox.com/app/account, to get to Settings.

> I'm not sure if there was a question here about whether there should be a settings icon in the main menu but if there was I think it is correct it is there as Settings is a part of the app just like Mail, Calendar etc. and it has a section of it's own just like the others.

I submit that Settings is **not** a part of the app just like Mail, Calendar etc -- rather that it is conceptually a meta-part as it controls how the other areas function (similar to Control Panel in Windows or System Preference on macOS), which was part of the initial reason for gathering all settings in one place. :)

> I don't think there is much to get confused about here once you've clicked on the Settings icon once (with or without a heading) :)

I think there is some room for confusion and that we should strive towards increased clarity especially where there might be a perception of redundancy or overlap, which I think is the case here.

Compare the following screenshot of the current Settings menu, while hovering over the Mail main menu item:
Runbox 7 Settings menu current

...with the proposed one:

Runbox 7 Settings menu proposed

In the first screenshot the main menu looks similar to the Settings menu, while in the second the difference is clearer.

Additionally, the purely design-related comments above are aesthetically relevant.

davidbowdley: Go with with your suggestion then. I have no strong feelings about this and just wanted to convey that I think once you've visited the section it will be clear how to use it. I was going to suggest that maybe there should be a different icon for mail settings on the left so that it isn't so similar to the one in the main menu but I don't think that is possible as we aren't using custom icons?

It just occurred to me that if you are not sure which section the option you are looking for is under an "expand all" option might be useful (different topic though entirely).

gtandersen:
I think the navigation issues will become clearer when Settings is shown as an overlay, and that we can temporarily keep the menu heading in the meantime.

gtandersen:
Merged in other changes and resolved conflicts, hopefully without losing anything.

tadzik (approved):

PR #997, opened 2021-05-17
fix(calendar): Display ical events with non-user timezones correctly

castaway:
Now displays everything in "timezone" setting from Account Details.


tadzik (approved):
Looks sensible – good stuff!

PR #999, opened 2021-05-21
Runbox container overhaul

tadzik:
This largely re-does <runbox-container> so that it fits our usage. As a proof of concept, I've ported the account-app to it, to see that it both fits existing components and massively simplifies things by reducing the (usually copypasted) sidebar implementations.

It's also an attempt at more consistency. In the spirit of that, the sidebar containing the hamburger button that hides the sidebar is made always visible (not just in mobile mode as before), so that we don't have some components that have it on because they need it (like calendar and webmail) and some that only show it (empty) in mobile view.

This of course begs the question of what should we put there? Webmail has a search bar, and perhaps that's an idea worth following (especially in big, complex components like the settings)?

This also brings back the discussion of "what do we put in the sidebar if we don't have anything to put there"?


gtandersen:
Nicely done @tadzik!

Agree that more consistency is needed, and that search would be useful in all areas of the app including Settings (this obviously requires some separate indexing and search utility).

We can always find something useful to put in the sidebar, and it being present across the app will help us structure the navigation more consistently.

hernan604 (approved):

kvviecien (approved):

PR #1000, opened 2021-05-21
Fix the build, make the changelog generator smarter

tadzik:
This unblocks the strict compilation (again, no idea why Github CI tends to not catch this?) and also makes the changelog builder cope with (Github's, I think) squashed branch merge commits.


gtandersen (approved):
Great @tadzik!

PR #1008, opened 2021-05-31
style(overview): Change default sorting and unread selectors, and imp…

gtandersen:
…rove their formatting.

Improved Overview selectors


tadzik (approved):
Looks good :)

PR #1011, opened 2021-06-03
fix(mailview): Ensure we invalidate the message cache for HTML update

castaway:


tadzik (approved):
Looks all good to me :)

PR #1016, opened 2021-06-08
fix(identitys): show the "delete" button for identitys except for "ma…

hernan604:
…in" & "aliases"


tadzik (approved):
Looks good to me :)

PR #1017, opened 2021-06-09
fix(mailviewer): Show inline HTML images again.

castaway:
Also don't list attachments for inline HTML images.


tadzik (approved):
LGTM

castaway:
-> Draft, as dependent changes have been reverted

PR #1019, opened 2021-06-13
style(welcome): Improve layout and make it more consistent.

gtandersen:


gtandersen:
Redone Welcome screen with hover effects, while making it more consistent with Settings and Help screens:

Runbox 7 Welcome screen

PR #1022, opened 2021-06-15
Fix unnamed contacts

tadzik:
This fixes some bugs reported by @davidbowdley, starting from unnamed contacts being filtered out by search (fixed in 81d4a253270ed886d7a2696598fa5e40db471102), followed up by other things uncovered by that.

The user-facing change is that creating contacts with no deducible name is no longer disallowed – explained more in depth in the [commit](https://github.com/runbox/runbox7/commit/89d3364c072febbb1e938fa3cb5572341eb8cb07) that changed it.

### DIFF
--- a/src/app/contacts-app/contact-list.component.ts
+++ b/src/app/contacts-app/contact-list.component.ts
@@ -69,7 +69,7 @@ export class ContactListComponent {
 
             return c.categories.find(g => g === target);
         }).filter(c => {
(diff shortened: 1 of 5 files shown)


gtandersen (approved):
Looks sensible, @tadzik @davidbowdley.

castaway (approved):
Looks good

PR #1037, opened 2021-06-23
Split Security into separate components, make dedicated Security menu

### LINKED ISSUE #989: Impossible to dismiss the Account Security password pop-up
It is impossible to dismiss the Account Security password pop-up and return to the interface. You can only do this by editing the URL in the browser.

### LINKED ISSUE #991: Accessing Account Security from Settings
If you click on Settings and scroll down to Account Security and click on that option, the page behind the password pop-up loads but doesn't scroll to the top. After entering your password and the pop-up goes away this can be a bit confusing as you can't see the tabs at the top and in any case it makes sense for the page to start at the beginning.

kvviecien:
<kbd></kbd>

Fixes #989 - It's now possible to close the security password dialog on ESC click
Partially fixes #991 - Tabs are removed, so it's more accessible but I can't get the page to stroll to top on route change


gtandersen:
https://user-images.githubusercontent.com/1620058/123158630-e589fe00-d420-11eb-8ca9-f2d88cbea3de.mov

tadzik (approved):

PR #1041, opened 2021-06-30
fix(settings): Fix position of Settings Menu title for mobile

kvviecien:
Before / After
<kbd></kbd><kbd></kbd>


gtandersen (approved):
Looks good @kvviecien!

PR #1048, opened 2021-07-12
feat(account-security): Add possibility to change Password

kvviecien:
Screenshots:

[Mobile view](https://i.imgur.com/uivXByc.png) / [mobile view with error message](https://i.imgur.com/ROgSyoy.png)
[Desktop view](https://i.imgur.com/nz0MmJ2.png) / [desktop view with error message](https://i.imgur.com/GEc0Vvz.png)  

---
Cases checked before submitting the form:
- If any of the fields are empty
- If new + confirm new passwords match
- If password is 8-64 characters long
- If contains at least 1 alphabetical character
- If contains at least 1 numeric character
- If contains at least 1 special character

The rest is checked by the backend (e.g. if old password is correct).

### DIFF
--- a/src/app/account-security/account-password.component.ts
+++ b/src/app/account-security/account-password.component.ts
@@ -1,0 +1,119 @@
+// --------- BEGIN RUNBOX LICENSE ---------
+// Copyright (C) 2016-2018 Runbox Solutions AS (runbox.com).
+//
+// This file is part of Runbox 7.
+//
+// Runbox 7 is free software: You can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the
+// Free Software Foundation, either version 3 of the License, or (at your
+// option) any later version.
+//
+// Runbox 7 is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
+// ---------- END RUNBOX LICENSE ----------
+import { Component } from '@angular/core';
+import { MatSnackBar } from '@angular/material/snack-bar';
+import { RMM } from '../rmm';
+import { share, timeout } from 'rxjs/operators';
+
+@Component({
+    selector: 'app-account-password',
+    styleUrls: ['account.security.component.scss'],
+    templateUrl: 'account-password.component.html',
+})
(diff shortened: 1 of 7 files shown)


gtandersen (approved):
Looks great @kvviecien!

src/app/account-security/account-password.component.ts
```diff
@@ -0,0 +1,94 @@
...
+    confirm_password: string;
+    error: string;
+    is_busy: boolean;
+
+    constructor(public snackBar: MatSnackBar, public rmm: RMM) {
+        this.rmm.me.load();
+    }
+
+    check_password(old_password = this.old_password, new_password = this.new_password, confirm_password = this.confirm_password) {
+        if (old_password && new_password && confirm_password) {
```
tadzik, line 41: This would be easier to read (and extend) with a "return early" pattern, like so:

```
if (!old_password || !new_password || !confirm_password) {
    this.error = 'Fields can not be empty';
    return;
}

if (new_password !== confirm_password) {
    this.error = "Passwords don't match";
    return;
}

...

// if we got this far down, the password is good – no need for elses
this.error = undefined;
this.update_password(old_password, new_password, confirm_password);
```

src/app/account-security/account-password.component.ts
```diff
@@ -0,0 +1,94 @@
...
+            } else {
+                this.error = "Passwords don't match";
+            }
+        } else {
+            this.error = 'Fields can not be empty';
+        }
+    }
+
+    update_password(old_password: string, new_password: string, confirm_password: string) {
+        this.is_busy = true;
```
tadzik, line 70: This isn't used anywhere :)

src/app/account-security/account.security.module.ts
```diff
@@ -51,16 +51,19 @@ import { ManageServicesComponent } from './manage-services.component';
...
     declarations: [
       ModalUnlockcodeComponent,
       ModalPasswordComponent,
+      AccountPasswordComponent,
       TwoFactorAuthenticationComponent,
       ManageServicesComponent,
       AppPasswordsComponent,
       LastLoginsComponent,
       SessionsComponent,
+      AccountPasswordComponent,
```
tadzik, line 66: Looks like a duplicate.

tadzik (commented):
Looks good; added some nitpicks to the code.

If we wanted to be more fancy (and a bit user-friendly probably), we could utilize [Angular's form validation](https://angular.io/guide/form-validation#validating-input-in-reactive-forms), that'd be able to display potential password requirement failures instantly, rather than waiting for the "Save" button to be pressed. Or we could accomplish the same by calling `check_password()` on a `change` signal on the password input field :)

PR #1051, opened 2021-07-14
fix(maillist): Ensure actions involving 2 folders update counts on both

castaway:
Fixes issues with displayed folder counts being incorrect after message moves (including to spam/trash) - intermittent issue that happens when a move is made when an index sync is also in progress

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -608,7 +608,9 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
         }
       },
       updateRemote: (msgIds: number[]) => {
-        const res = this.rmmapi.trainSpam({is_spam: params.is_spam, messages: messageIds});
+        const userFolders = this.messagelistservice.folderListSubject.value;
+        const currentFolderId = userFolders.find(fld => fld.folderName === this.messagelistservice.currentFolder).folderId;
+        const res = this.rmmapi.trainSpam({is_spam: params.is_spam, from_folder_id: currentFolderId, messages: messageIds});
         res.subscribe(data => {
           if ( data.status === 'error' ) {
             snackBarRef.dismiss();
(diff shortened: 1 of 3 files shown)


src/app/app.component.ts
```diff
@@ -917,6 +921,10 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
   dropToFolder(folderId): void {
     const messageIds = this.canvastable.rows.selectedMessageIds();
 
+    let currentFolderId;
+    this.messagelistservice.folderListSubject.subscribe(folders => {
```
tadzik, line 925: I think a race is possible here: messageActionsHandler should probably wait for this call to finish before it fires its own.

tadzik (changes requested):

src/app/app.component.ts
castaway: folderListSubject is a BehavorSubject, which is documented as returning its current value immediately..
we already use the exact same construction in existing code ... so I think its fine?

castaway: Moved the variable closer to its use, and switched to behaviorsubject .value, which we don't have to wait on. (If user's are running actions against messages/folders, and don't have folders, something is quite broken..)

src/app/app.component.ts
```diff
@@ -963,8 +975,12 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
             this.clearSelection();
             this.canvastable.rows.clearOpenedRow();
           },
-          updateRemote: (msgIds: number[]) =>
-            this.messagelistservice.rmmapi.moveToFolder(msgIds, folder)
+          updateRemote: (msgIds: number[]) => {
+            console.log('update remote: start');
```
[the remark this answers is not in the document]
castaway: fixed

tadzik (approved):
Looks good ~~– assuming that the logs are staying intentionally~~ :)

hernan604 (approved):

PR #1053, opened 2021-07-28
Ensure user action results (index updates) take precedence over generic background updates

castaway:
searchservice's regular update has been amended so that if called with different arguments, from various "updateMessages" calls, it will fetch the given updates next - these arguments contain a separate "last changed time" which is extracted from the REST response to the action (eg moveMessages will return the earliest (updated) timestamp of the items moved).

After the interrupt update, the regular updates resume (the interruption does not change the normal "last update time").

If the system is already fetching the new changes, the interruption update does not duplicate the action (we now store which messages the index update is working on, so that we can compare them).

This PR also contains a commit which ensures that when fetching message data results in an error (issue with the data on the backend), it is appropriately dealt with and does not interrupt/kill anything else.


gtandersen (approved):
Great work @castaway!

hernan604 (approved):

PR #1055, opened 2021-08-02
fix(settings): Add missing Domain Registration tile

kvviecien:


gtandersen (approved):
Good catch @kvviecien!

PR #1058, opened 2021-08-05
fix(mailviewer): Update maillist after delete/move

castaway:
The main tweak here is that after updating the new set of rows (MessageDisplay content in the canvastable) we force an update of the canvastable "topindex", which will have changed when messages are removed from the current folder. Why exactly this works fine in other circumstances.... I'm not sure!

Fixes blank-list issue


gtandersen (approved):
Excellent fix @castaway.

PR #1071, opened 2021-09-03
Castaway/message load errors

### LINKED ISSUE #1067: Show error message with link to Service Status when a message can't be opened
When a message can't be opened (except in the case of network issues which is handled separately), show an error message with a link to Service Status similar to RMM6 (but replace the Compose link with a link that works in Runbox 7):

Error: The message could not be opened. Please see <a href="https://status.runbox.com/" class="activeerror" target="status">Service Status</a> for possible server problems, or contact Runbox Support at <a href="/mail/compose?to=support@runbox.com" class="activeerror" target="compose">support@runbox.com</a>.

castaway:
Fixes #1067 among other things.

Splits existing "errors" into "warnings" and "errors" - an error is when the backend had perl errors and couldn't actually complete the request, a warning is when the backend completed, but couldn't do what was asked. For warnings we can continue, and display the user a reason, and a prompt to contact support.

This will require a backend change to determine which errors should actually be warnings.

### DIFF
--- a/e2e/cypress/integration/mailviewer.ts
+++ b/e2e/cypress/integration/mailviewer.ts
@@ -7,12 +7,24 @@ describe('Interacting with mailviewer', () => {
 
     beforeEach(() => {
         localStorage.setItem('localSearchPromptDisplayed221', 'true');
+        localStorage.setItem('messageSubjectDragTipShown', 'true');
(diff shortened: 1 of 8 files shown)


gtandersen (approved):
Nicely done @castaway!

PR #1073, opened 2021-09-08
fix(mailviewer): Remove big icons from print view

### LINKED ISSUE #889: Print/save as PDF includes message header icons
When clicking the Print icon in a message and choosing "Save as PDF", the resulting PDF includes the icons shown in the header (in a very large size):

Print example

kvviecien:
Fixes #889


gtandersen (approved):
Looks good @kvviecien :)

gtandersen:
Most of the icons are gone now, but it seems a couple remain:
Runbox 7 print view icons

Could you have another look @kvviecien? :)

gtandersen:
They are visible in Firefox on macOS.

Issue #1033, opened 2021-06-19
Improve header area of messages


gtandersen:
When displaying messages the header area looks disorganized, and we should:

- [x] Adjust margins around header entries
- [x] Remove dotted lines under contacts that don't exist in Contacts and adjust the "Add contacts" icon size
- [x] Replace the icon for "Show basic headers" (should be `unfold-more-horizontal`)

gtandersen:
Nicely done, @piratefinn!

PR #1074, opened 2021-09-09
Piratefinn/gh 1033/mailheader improvements

piratefinn:
Addresses #1033 
Contact icon scaled down to 0.7 with transform on contactcard, icon resized to fit more with height of text and not overwhelm it and icon changed where relevant.

### DIFF
--- a/src/app/mailviewer/contactcard.component.ts
+++ b/src/app/mailviewer/contactcard.component.ts
@@ -28,7 +28,7 @@ import { AppSettingsService } from '../app-settings';
     // tslint:disable-next-line:component-selector
     selector: 'rmm7-contact-card',
     template: `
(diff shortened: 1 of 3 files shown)


gtandersen (approved):
Looks good @piratefinn!

gtandersen:
It would be even better if the subject line had the same left margin/padding as the other headers. :)
Runbox 7 message headers

gtandersen:
And the expanded headers could benefit from the same indent:
Runbox 7 message headers expanded

piratefinn:
> It would be even better if the subject line had the same left margin/padding as the other headers. :) Runbox 7 message headers

This one I was unsure on when I was initially adding the padding because I liked how it looked padded in line with the content, but if you wish it changed then I'll do that. The latter I'll do now.

gtandersen:
Ideally all the headers and the content would have the same padding, perhaps?

Issue #1078, opened 2021-09-11
Always show message view options


gtandersen:
The message view options menu items depend on the use of the local index.

**With index:**
<img width="274" alt="Message view options with index" src="https://user-images.githubusercontent.com/1620058/132955172-91711a80-6c5e-42a2-bb58-941eeeb35578.png">

**Without index:**
<img width="276" alt="Message view options without index" src="https://user-images.githubusercontent.com/1620058/132955171-5dbf507f-e2d3-4116-9f3c-1ff6583468d5.png">

This can be confusing and we could instead show all options regardless, but when the local index is not in use gray out disabled options with a tooltip explaining that these options are only available when the index is in use.

piratefinn:
@gtandersen confirm this is fixed with [this PR](https://github.com/runbox/runbox7/issues/1078) ?

gtandersen:
@piratefinn Thanks; confirmed and closing issue.

PR #1081, opened 2021-09-15
Castaway/979 preview unread filter non index

### LINKED ISSUE #959: Allow "unread only" mode to also work when the index isnt synchronised
The data is there, I know it is.. Also its a little jarring clicking over to the Spam folder (which isnt indexed) and seeing unread messages

castaway:
* Add "plain text" email source into the "messagelist" objects which we use for displaying emails when we're not using the xapian index.
* Use the above data to enable "inline previews" when index is not synced
* Filter the "messagelist" rows by seen/not seen, to enable "Unread only" when index is not synced
* "Threaded view" not available with no index - so this is greyed out/disabled if the index is not synced 

Fixes #959 and #1078

### DIFF
--- a/src/app/common/messagelist.ts
+++ b/src/app/common/messagelist.ts
@@ -89,7 +89,12 @@ export class MessageList extends MessageDisplay {
         name: 'Subject',
         sortColumn: null,
         getValue: (rowIndex: number): string => this.getRow(rowIndex).subject,
-        draggable: true
+        draggable: true,
+        getContentPreviewText: (rowIndex): string => {
+          const ret = this.getRow(rowIndex).plaintext;
+          return ret ? ret.trim() : '';
+        },
+        // tooltipText: 'Tip: Drag subject to a folder to move message(s)'
       },
       {
         sortColumn: null,
--- a/src/app/common/messagelist.ts
+++ b/src/app/common/messagelist.ts
@@ -62,6 +62,14 @@ export class MessageList extends MessageDisplay {
     '';
   }
 
+  // filter visible rows by whatever options the frontend has
+  filterBy(options: Map<String, any>) {
+    this.rows = this._rows;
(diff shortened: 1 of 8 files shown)


src/app/common/messagelist.ts
```diff
@@ -89,7 +97,12 @@ export class MessageList extends MessageDisplay {
         name: 'Subject',
         sortColumn: null,
         getValue: (rowIndex: number): string => this.getRow(rowIndex).subject,
-        draggable: true
+        draggable: true,
+        getContentPreviewText: (rowIndex): string => {
+          const ret = this.getRow(rowIndex).plaintext;
+          return ret ? ret.trim() : '';
+        },
+        // tooltipText: 'Tip: Drag subject to a folder to move message(s)'
```
kvviecien, line 105: Should this be uncommented?

kvviecien (approved):
One more thing that could be improved would be adding a tip for the disabled options that the user should enable Index to use them?

Right now it looks like this:
<kbd></kbd>
And instead it could say something like "Synchronise index to use this view"

gtandersen:
Agreed @kvviecien. :)

src/app/common/messagelist.ts
castaway: It was there when I started.. dunno why its commented ;)

castaway:
> One more thing that could be improved would be adding a tip for the disabled options that the user should enable Index to use them?
> 
> Right now it looks like this:
> 
> And instead it could say something like "Synchronise index to use this view"

I'm not sure how to add a "tooltip when disabled" - any ideas?

src/app/common/messagelist.ts
gtandersen: I suspect ti's been commented out because a similar message is shown in the snackbar when using Runbox 7 for the first time. :)

PR #1101, opened 2021-10-12
Fixing padding

piratefinn:
Based on feedback in #1074 to bring everything more in line and tidy, such that header padding aligns with content and applies to extended headers.


castaway:
Does changing the styling  from % to explicit px have an affect on other layouts? (mobile etc)

PR #1115, opened 2021-11-02
Castaway/1083 folder col width

castaway:
Allow the folder column to be resized on the right end - this is done by adding an empty/zero width column on the end (could not see how to add it otherwise!)

Also store the user's "saved column widths" in sets, so if the (search results) view including the folder column is resized, these sizes are not applied when viewing the normal message list.

Imports previously saved column width info, though this is unlikely to be usable, so users may need to recreate any saved widths they had.


gtandersen (approved):
Great work @castaway!

Issue #883, opened 2021-02-04
Attachment far below the message body


kotecky:
Since a long time (or since ever?) and very often, but not always, the attachment tiles render far below the message body, with a white space in between (FF on Win10; not in FF on Android). It keeps causing the "did they forget the attachment?" question. And the need to scroll and scroll and scroll.

castaway:
Looking at attachments, even without a big gap (which I havent yet seen), I think we could do better to display them more obviously - maybe move attachments to the top of the email (with a collapsible title), or add a link that jumps to the attachments in the mail view menu?

gtandersen:
@castaway Agreed; it would be intuitive to list the attachments near the top of the message. Showing previews is also useful, so we could do both.

castaway:
Not sure what you mean by previews in this context?

gtandersen:
@castaway When the attached file is an image, a preview/thumbnail is shown in the attachment area:
<img width="240" alt="Screen Shot 2021-08-11 at 20 14 45" src="https://user-images.githubusercontent.com/1620058/129081702-f9d0df27-fe0e-4a63-9d84-35e9d9a45c2d.png">

castaway:
Shuffling things around a bit and making the attachments collapsible:

Default (closed):
![image](https://user-images.githubusercontent.com/109990/139847975-8baee0b5-b138-436c-b1cf-8ff5e1461ef5.png)

Open:
![image](https://user-images.githubusercontent.com/109990/139848048-763b0c7e-6fd1-4720-8b55-684e994d19a6.png)

I'm thinking the "Show attachments" title could be a row of attachment icons instead?

gtandersen:
Looks good @castaway!

Perhaps the Show Attachments drop-down would be more visible with some color (if only just a blue font color), so that users more easily understand that it's clickable?

gtandersen:
Or as you suggested further down, a row of attachment icons would be neat!

kotecky:
Observer's remarks: 
* whether attachments are displayed or hidden should be a setting, not a behavior. Ideally per device (as different setting needs can be expected on a phone and on a computer. 
* the icons/previews should be much smaller.
* previews should be framed or have a BG color (if blank?)  (some are empty and so invisible)
*  something (icon? text?) should tell the user the type (PDF, image, doc,...) 
* attachment file size is needed. Preferably 3.141 MB (or at least 3 141 kB with a separator every 3 digits) rather than 3141592.

[...]

PR #1119, opened 2021-11-03
refactor(mailviewer): Display a list of attachments at the top

castaway:
Add an expandable list of attachments near the top of the email view. Needs some tweaking (please @kvviecien - how do we get the horizontal divider?)

Fixes: #883

### DIFF
--- a/src/app/mailviewer/mailviewer.module.ts
+++ b/src/app/mailviewer/mailviewer.module.ts
@@ -26,6 +26,7 @@ import { MatDialogModule } from '@angular/material/dialog';
 import { MatDividerModule } from '@angular/material/divider';
 import { MatExpansionModule } from '@angular/material/expansion';
 import { MatGridListModule } from '@angular/material/grid-list';
+import { MatListModule } from '@angular/material/list';
 import { MatIconModule } from '@angular/material/icon';
 import { MatMenuModule } from '@angular/material/menu';
 import { MatRadioModule } from '@angular/material/radio';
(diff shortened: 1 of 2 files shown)


castaway:
image

Hmm need some more icons..

gtandersen:
Nice work @castaway! 

How about one of these for graphics:

https://materialdesignicons.com/icon/file-image
https://materialdesignicons.com/icon/file-image-outline

castaway:
The missing icons on mine are due to backend issues, the attachments extractor uses nodejs mailparser, which doesnt seem to be running on rmm604 ... I should probably switch to 704 ..

kvviecien:
@castaway you can create a divider by using `<mat-divider></mat-divider>` from Angular Material. Is that what you need?

gtandersen:
Looking forward to seeing this in action @castaway. :)

castaway:
With divider:
Closed: 
image

Open:
image

Having both looks a little silly to me in a rather short email.. I assume less so in a long one .. 

Current thoughts: 

* I have it set to "default open, unless mobile" - I think I prefer default closed, regardless..
* Should we make the bottom-of-page list closeable too, or is that just my terrible aesthetic sense?

Parser working now - still missing some file types / a default? need to poke mailparser separately

gtandersen:
Nicely done @castaway!

We could perhaps list the attachments on one line at the top of the message (with or without the small previews), and then the larger versions could ideally be shown near the bottom of the visible window.

Would it be possible to show somewhat larger icons in the bottom preview row? :)

PR #1120, opened 2021-11-03
FIX for #235 "Reply header" missing separators or spaces

### LINKED ISSUE #235: "Reply header" missing separators or spaces
The "reply header" (from, date, to, subject) displayed on top of the text of the message being replied to, misses separators or spaces. Example: 

`From: Info <info@abc.org>Time:2019-08-14 11:24To: Peter <peter@abc.org>RE: topics`

kvviecien:
Fixes #235 

Created a reply header for HTML messages. It didn't exist, we used the same header for everything. Now it looks like this:
<kbd></kbd>

Improved header for forwarding HTML messages, looks like this:
<kbd></kbd>

### DIFF
--- a/src/app/compose/draftdesk.service.ts
+++ b/src/app/compose/draftdesk.service.ts
@@ -162,7 +162,7 @@ export class DraftFormModel {
             ret.html =
                 `<br />
                 <hr style="width: 100%" />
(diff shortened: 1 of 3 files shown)


gtandersen (approved):
Looks great @kvviecien!

PR #1127, opened 2021-11-18
fix(mailviewer): Fix horizontal scroll for iframe in message view

### LINKED ISSUE #1072: Can't scroll horizontally in message view
When viewing an HTML or preformatted message in vertical mode (Firefox on macOS), the horizontal scrollbar isn't visible and it's not possible to scroll horizontally.

kvviecien:
Fixing again. Works better than last time. 
HTML messages now have a scrollbar and iFrame doesn't seem to break, so hopefully I nailed it this time.

Fixes #1072


gtandersen (approved):
Excellent!

PR #1144, opened 2021-12-03
fix(login): Complete rest of status error messages.

gtandersen:
The status error messages are now as follows:

**Status 1**
Status 1
**Status 2**
Status 2
**Status 3**
Status 3
**Status 4**
Status 4
**Status 5**
Status 5
**Status 6 and up**
Status 6

### DIFF
--- a/src/app/login/login.component.html
+++ b/src/app/login/login.component.html
@@ -46,6 +46,17 @@
           Your Runbox subscription has expired. <br>
           Go to <a routerLink="/account/subscriptions">Your Subscriptions</a> to renew it now!
 	</p>
+	<p *ngIf='accountCanceled' class='login_error_html'>
+          Your Runbox account has been canceled. <br>
+          Please contact <a href="https://support.runbox.com">Runbox Support</a>.
+	</p>
+	<p *ngIf='accountClosed' class='login_error_html'>
+          This Runbox account has been closed. <br>
+	</p>
+	<p *ngIf='accountError' class='login_error_html'>
+          An unknown error has occurred. <br>
+	  Please contact <a href="https://support.runbox.com">Runbox Support</a>.
+	</p>
 	<div id="createAccount" class="login-link">
 	  <a href="/signup?runbox7=1">Create new account</a>
   </div>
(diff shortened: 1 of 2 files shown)


gtandersen:
Does this look OK @runboxdave?

runboxdave:
I was going to say that canceled has two lls unless you are in the US (which we are not) but actually it's not really a useful word because it kind of implies something different to it being closed when in fact it is just closed really. So maybe just used closed for Status 5 and up?

hernan604 (approved):

hernan604:
@gtandersen its failing the lint tests. try to run: 

npm run lint        #should display where are the errors

npm run lint -- --fix     #it will try to autofix

gtandersen:
@runboxdave The distinction was made because status 5 results from the customer initiating the closure (as in cancellation of the subscription), while status > 5 implies closure by staff regardless of any subscription.

gtandersen:
@hernan604 Thanks; done now.

runboxdave:
> @runboxdave The distinction was made because status 5 results from the customer initiating the closure (as in cancellation of the subscription), while status > 5 implies closure by staff regardless of any subscription.

Yes, I realise why the distinction is made. However, "cancelled" doesn't actually indicate what is intended (it could be cancelled by us and sounds somewhat brutal). Perhaps the response can be "Closed by customer" instead (or "cancelled by customer" if you prefer that)?

gtandersen:
@runboxdave Yes, I see the point; however space is limited and for consistency I have changed the error message to "Your Runbox subscription has been canceled." for status 5 now.

For status > 5 the error is "This Runbox account has been closed.", such that a distinction is made between "subscription" and "account" in addition to "canceled" and "closed".

Hopefully this clarifies it sufficiently; otherwise we can make further improvements later.

runboxdave:
Note that the status of the account (at least for Status 3 and 4) is disclosed to anyone trying to log in with a correct username but incorrect password. Given that usernames are generally public (i.e. they are email addresses) we are giving away account status when we shouldn't be.

gtandersen:
@runboxdave Please open a new issue for this. :)

gtandersen:
Already done in internal issue #1149.

Issue #234, opened 2019-08-15
"Undefined" ipo header when replying


kotecky:
When replying to a message, the word "undefined" is placed in place of the "reply header" (from, time, to, subject). Spotted once, could not reproduce even with the same message, message being replied-to has all fields (from and to email, subject not blank).

> ![image](https://user-images.githubusercontent.com/2297444/63085849-ce3d7800-bf4e-11e9-940c-1726d9311e6d.png)

gtandersen:
Not able to reproduce this. Closing issue; reopen if it occurs again.

kotecky:
Since recently i get it every time, irrespective of the sender's client (desktop, phone app, automated email). HTML email seems to be a precondition. 

( Runbox 7 build time: 2021-11-15T13:08:33.650Z )

gtandersen:
Yes, I see this too now on HTML messages.

@kvviecien Could this be related to https://github.com/runbox/runbox7/pull/1120, or is it independent?

kvviecien:
@gtandersen There is a possibility my fix could break after fixing the headers for recent Sentry issues, because that's the same part of the code. I will look into it again.

gtandersen:
This appears to be much better but can still be reproduced with some messages, e.g. with https://runbox.com/app/#Inbox:12205016 in the test account.

kotecky:
Still there, confirmed.

kotecky:
Is it not persistent when answering to more than one recipient?

gtandersen:
@kotecky Thanks for the updates. 

Do you mean that the error occurs when replying to one recipient but not when replying to more than one recipient?

kotecky:
@gtandersen, that's what I meant, but dud not try. A hypothesis to be tested.

gtandersen:
Not reproducible with the above example message.

Closing issue; reopen if it occurs again.

kotecky:
@gtandersen To be reopened as it is currently happening again.

kotecky:
> @gtandersen To be reopened as it is currently happening again.
@gtandersen Issue still there.

gtandersen:
@kotecky Issue reopened.

Would you be able to provide an example message (your Runbox username and the message's ID from the browser URL) via Runbox Support?

Thanks!

kotecky:
@gtandersen 
Answering to the support via a ticket.

kotecky:
[...]

PR #1148, opened 2021-12-08
fix(mailviewer): Properly show reply and forward headers for HTML messages

kvviecien:
Improves: #234 
Fixes sentry errors about this part of code.

Unfortunately the header still rarely renders as undefined. This fix does improve this issue a lot though. 
I'm working on figuring out why this is happening.

### DIFF
--- a/src/app/mailviewer/singlemailviewer.component.ts
+++ b/src/app/mailviewer/singlemailviewer.component.ts
@@ -97,10 +97,10 @@ export class SingleMailViewerComponent implements OnInit, DoCheck, AfterViewInit
   @ViewChild('messageContents') messageContents: ElementRef;
   @ViewChild('htmliframe') htmliframe: ElementRef;
   @ViewChild('htmlToggleButton') htmlToggleButton: MatButtonToggle;
-  @ViewChild('replyMessageHeader') replyHeaderHTML: ElementRef;
-  @ViewChildren('replyMessageHeader') replyHeaderHTMLQuery: QueryList<ElementRef>;
-  @ViewChild('forwardMessageHeader') messageHeaderHTML: ElementRef;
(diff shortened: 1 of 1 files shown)


src/app/mailviewer/singlemailviewer.component.ts
```diff
@@ -218,13 +218,21 @@ export class SingleMailViewerComponent implements OnInit, DoCheck, AfterViewInit
...
+          this.mailObj.origMailHeaderText = messageHeaderHTMLList.first.nativeElement.innerText;
+        }
+      }, 0);
     });
 
     this.replyHeaderHTMLQuery.changes.subscribe((replyHeaderHTMLList: QueryList<ElementRef>) => {
-      this.mailObj.origReplyHeaderHTML = replyHeaderHTMLList.first.nativeElement.innerHTML;
-      this.mailObj.origReplyHeaderText = replyHeaderHTMLList.first.nativeElement.innerText;
+      setTimeout(() => {
+        if (this.messageHeaderHTML) {
```
castaway, line 231: Shouldn't this one be `if(this.replyHeaderHTML)` ?

castaway (commented):
Needs the typo fixing - also my tests show it works for me just as well without the setTimeout part (just need the if(this .. ) - so I think that does nothing?

kvviecien:
Unfortunately I couldn't find a better way to tackle this. When I was trying not to use the template and did everything in JS, it still ended up undefined.

So this is the best solution from me. Most of the time it defines the header properly, but if you switch emails really fast and immediately click reply, it will probably end up undefined. It's still an improvement from what's happening now. 

Do you want to give it a try @castaway ?

gtandersen (approved):

PR #1168, opened 2022-01-10
Fix default orientation when not set on narrow width

### LINKED ISSUE #1162: Use horizontal preview by default on small screens
**Describe the bug**
On small screens such as smartphones it is not useful with the default vertical message preview pane, and it should instead default to horizontal preview.

**To Reproduce**
1. Open the PWA (app) or go to https://runbox.com/app with a smartphone and log in
2. Tap to open a message
3. Observe that the default behavior is a vertical preview pane (unless changed manually)

**Expected behavior**
The default preview pane mode should be horizontal on small screens, so that the message list is shown at the top and the preview underneath. 

The preview pane can then be expanded to fill the screen by using the up-arrow in the message action menu.

**Screenshots**
IMG_0EEEE8FF3F93-1

 - OS: iOS
 - Browser: Safari

piratefinn:
Fixes #1162 by checking width properly before setting if to put on the single mail view on side or bottom before it is set and saved in localStorage as before. Also uses the newer API for width check, making it one of the first to use it very demonstrably in app flow. Could be possible to tidy up to look nicer/more concise, but regardless it works from personal testing. Test with another platform such as Mac first ( @gtandersen ?) before merge though.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -356,6 +356,7 @@
 <router-outlet (activate)="childRouteActivated(true)" (deactivate)="childRouteActivated(false)"></router-outlet>
 </mat-sidenav-container>
 
+<!-- This will cause an error in dev mode with the async pipe, should not be present however in production -->
 <div style="position: absolute; bottom: 2px; right: 2px; z-index: 10000;"
     *ngIf="progressService.httpRequestInProgress | async"
     >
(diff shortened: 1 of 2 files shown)


gtandersen:
Thanks @piratefinn -- @hernan604 @runboxdave we can test this in staging once all the CI tests pass.

piratefinn:
> Thanks @piratefinn -- @hernan604 @runboxdave we can test this in staging once all the CI tests pass.

Sounds good. The CI test seems to be on something unrelated but I assume you mean "when we fix CI tests first we'll stick this into staging".

castaway (approved):
Neat, and tests are now passing it looks like - lots of other places we use ".matches" on mobile, should probably review all those too?

piratefinn:
> Neat, and tests are now passing it looks like - lots of other places we use ".matches" on mobile, should probably review all those too?

They're simply the older API which has yet to be updated, but an issue could be made of I assume a lower priority to update all those to use the newer one.

PR #1188, opened 2022-01-24
style(mail): Replace "mark unread" icon.

gtandersen:
Unread

Ref: https://community.runbox.com/t/mark-read-mark-unread/1648

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -253,7 +253,7 @@
                 [disabled]="viewmode==='conversations'"
                 class="tableViewOptionsMenuElement"
               >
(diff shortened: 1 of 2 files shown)


hernan604 (approved):

kotecky:
Somewhat related to this is the button "move to folder". It is related by it's visual similarity of a blue rectangle - when working with emails I do not look much on the details on the icon and just click on the closest with the right appearance. Then a folder selection pop-up opens and i'm puzzled, not knowing what should be my next action, did i want to select a folder? (I was not, the mail was supposed to be marked as read.) This happens (to me) rather often.

A possible solution is an icon with an arrow showing that the file will go into a folder, like in the Fairmail app, fourth icon from left in the row of icons.

image

PR #1193, opened 2022-01-25
fix(mailview): Remove deleted/moved messages before mesg list redraw

### LINKED ISSUE #713: Deleting multiple messages resulted in an empty message list

castaway:
Fixes #713 (hopefully!)

The logic for delete/move messages was not removing the messages from
the separate "message display" list of ids that the canvas uses to
paint its list from. A regular repaint was trying to redraw the list
using the old list of ids, fetching data from the index which had
already been deleted, this errored out, quitting in the middle of
redrawing the message list.

This fix:
1) ensures we remove the items from the message display first
2) ensures the repaint continues even if it cant find some of the
message data.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -621,6 +621,8 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
       updateLocal: (msgIds: number[]) => {
         // Move to spam folder (delete from index), set spam flag
         if (params.is_spam) {
+          // remove from message display
+          this.canvastable.rows.removeMessages(messageIds);
           this.searchService.deleteMessages(msgIds);
           this.messagelistservice.moveMessages(msgIds, this.messagelistservice.spamFolderName, true);
         } else {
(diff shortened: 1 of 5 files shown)


castaway:
Set to draft for a discussion/decision on:

Should we change existing logic from "quit repainting the message list in the middle if one of the messages has an issue" to "skip over erroring messages and continue" (the underlying issue is fixed, there could be others though)

gtandersen:
@castaway Yes since the underlying known issues is resolved, let's switch to skipping failing messages and log output to the console so we can identify any further underlying causes.

castaway:
Releasing this from Draft to Review as it currently does as Geir suggested. (With a minor caveat, we will get a white/blank row where the missing message would be)

gtandersen:
Nice work @castaway; will be interesting to see it in action. :)

gtandersen (approved):

gtandersen:
@castaway The list now displays a blank entry after deleting a message, where the deleted message was shown. In smaller accounts the empty entry disappears after a short while, but in larger accounts it seems to remain:
Message list empty entry

PR #1198, opened 2022-02-07
Cope with far future dates, folder separators always a "."

castaway:
1. Messages with future message dates would cause the index updating to stop (as it was then fetching "messages changed after 2038" for example) - fixed by ensuring message fetching compares change dates no later than the current time.

2. Sub-Folder paths were in some places stored/used with a "/" separator, and in some with a "." As "." is officially used, and we don't display the names (except for in the url when bookmarking a folder), this is changed to be "." everywhere. This was likely causing folder count errors (storing counts as "." path, but reading out as "/" path).

3. Fixed tests to correctly send message times in seconds, not milli seconds (they broke after implementing no1)


gtandersen (approved):

gtandersen (approved):
Nice work @castaway!

PR #1213, opened 2022-02-17
Castaway/843 dont scroll to top

castaway:
Fixes a bug introduced inadvertently yesterday (and refactors a variable name that was confusing!)


gtandersen (approved):
Thanks @castaway!

PR #1215, opened 2022-02-23
Castaway/turn index verification back on

castaway:
"verifymessages" a system to compare rest api + index folder counts at idle points (aka when no new index updates are happening), was broken in late 2020 - this fixes it. This will fix recent issues where the folder display counts are wildly incorrect until scrolling to the bottom of the folder. (papering over whatever the actual issue is, tho)

Also tweaked, a few bits that showed up in tests (like endlessly rerunning code to find the text of a message thats erroring on the backend)


gtandersen:
Well done @castaway -- probably a good idea to log these discrepancies so that we can also investigate the underlying cause.

gtandersen (approved):

PR #1216, opened 2022-02-23
Fix Size and Folder columns issues

### LINKED ISSUE #1126: Size and Folder columns work inconsistently
**Describe the bug**
The Size and Folder columns work somewhat inconsistently especially in Safari after the improvements in #1115.

**To Reproduce**
1. Go to https://runbox.com/app
2. Ensure the local index is in use.
3. Observe the following, as seen in the screenshots below:
 - The Size column heading is offset relative to the column content.
 - After entering a search term to bring up the Folder column in the message list, some of the longer the folder names are truncated and the Folder column can't be expanded to view the entire contents.
 - Similarly, the Folder column heading can't be slid all the way to the left next to the Size and flag/attachment columns. 

**Expected behavior**
The columns and their headings should be more flexible to allow the fullest possible use of the horizontal space.

**Screenshots**

Runbox 7 Size column heading offset Safari

Runbox 7 truncated folder name

https://user-images.githubusercontent.com/1620058/141159511-91dda939-a5ce-4298-b4bd-72d3139968f4.mov

 - OS: macOS
 - Browser: Safari

kvviecien:
- Changed order of the Folder column to be able to freely resize the Size column. This way the flag columns aren't getting in the way.
- Fixed the Size column alignment. Now it's aligned to the left like everything else.

Close #1126

### DIFF
--- a/src/app/common/messagelist.ts
+++ b/src/app/common/messagelist.ts
@@ -108,7 +108,6 @@ export class MessageList extends MessageDisplay {
         sortColumn: null,
         name: 'Size',
         rowWrapModeHidden: true,
-        textAlign: 1,
         getValue: (rowIndex: number): number => this.getRow(rowIndex).size,
         getFormattedValue: MessageTableRowTool.formatBytes,
       },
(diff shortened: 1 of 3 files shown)


gtandersen (approved):
Looks good @kvviecien, and makes more sense too. :)

PR #1232, opened 2022-03-30
feature(mailviewer): Do not load external HTML images unless asked

### LINKED ISSUE #31: Provide an option for HTML without images
There appears to be no option to show HTML formatted emails without images. Without this option you have to use either HTML with images (which may be undesirable) or the plain text version of a message if it exists. 

The plain text version of a message intended to be see in HTML where possible can be difficult to navigate as the text will contain multiple image links making the actual message text difficult to read.

castaway:
Not loading external image src URLs means the HTML mail view loads
faster, and we also do not load "tracking" images embedded in emails.

Currently this displays a "show images" checkbox (next to the HTML one), which defaults to off and cannot be set to "always". Ideas on defaults/implementation welcome. (Personally am tempted to make it an "always allow for this sender" rather than global on/off)

Fixes #31


gtandersen:
Great @castaway; looking forward to testing it!

PR #1233, opened 2022-03-30
refactor(mailviewer): Added DOMPurify for html sanitising

castaway:
Sanitize HTML using a well-known and maintained library


gtandersen (approved):
Nice, @castaway!

PR #1239, opened 2022-04-12
Castaway/1234 incorrect flags

castaway:
Fixed issue with answered/flagged flag display being incorrect (moving the folder column earlier made the indexes change, and we were caching values based on the col index) - now has a dedicated cacheKey per column type. As a bonus this means all from/to values are cached together.

Tidied up some code too, canvas columns had some bits we were not using at all.

Also added caching for values as well as formatted values (as the values also are result of effort/calculations, often)


gtandersen (approved):
Excellent work @castaway!

Issue #1065, opened 2021-08-19
Certain emails don't open


VSirvio:
Certain kind of automatic emails refuse to open in Runbox 7. When clicked, they only show a blank preview pane. In Runbox 6, they work fine. See an example of such a message below:

[message.txt](https://github.com/runbox/runbox7/files/7013090/message.txt)

OS: Windows 10 Home
Browser: Firefox 91.0.1 (64-bit)
Local index: Yes

gtandersen:
See also #1166.

PR #1248, opened 2022-04-22
fix(mailviewer): Default dates to 1970 where not supplied

castaway:
RFC compliant messages *must* have a Date header, however now we've found one that doesn't - default it to 1970 so that we can display it anyway

Fixes: #1065


gtandersen (approved):
@castaway Looks good!

PR #1250, opened 2022-04-27
feature(mailviewer): Store "show html images" per sender

### LINKED ISSUE #1249: Make the new images/external content feature have a permanent option
Since introducing the new "Show images" feature for HTML messages it is now somewhat annoying this has to be selected for every individual message whereas the show HTML feature has an option for it to remain turned on for all messages.

Make "show images" work like the HTML feature so there is an option to turn it on permanently (and an option to turn this off too).

**Describe alternatives you've considered**
None

**Additional context**
None

castaway:
The recent new feature to only show external images (et al) when
requested, implemented to speed up HTML display, is annoying some users.

This improvement allows users to save the "show external html images"
per sender (stored as a separate contact type in the dav contacts).

A new contact is only created if they do not already exist.

If/when the user later creates a contact with the same email as the
hidden type, we re-use the same contact and change it to the
"individual" type so that it will be displayed in Contacts.

Fixes #1249

### DIFF
--- a/src/app/contacts-app/contact.spec.ts
+++ b/src/app/contacts-app/contact.spec.ts
@@ -301,4 +301,14 @@ END:VCARD`);
         sut = Contact.fromVcard(null, sut.vcard());
         expect(sut.photo).toBe(uri);
     });
+
+    it('can create a SETTINGSONLY contact', () => {
+        const sut = new Contact({});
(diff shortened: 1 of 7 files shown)


gtandersen (approved):
Excellent, @castaway!

PR #1252, opened 2022-05-03
feature(mailviewer): More control on how html/image settings are stored

### LINKED ISSUE #1249: Make the new images/external content feature have a permanent option

castaway:
Attempt #2: Allow choice of storing HTML Always/Per sender/with/without images

Fixes #1249


castaway:
image

image

image

(etc.) - the orange triangle now doesnt do anything.

PR #1265, opened 2022-05-26
Castaway/644 refresh drafts if changed

castaway:
Fixes a bug where any Drafts that were edited or sent from outside of Runbox7 (eg via IMAP or RMM6)  were not reflected in Runbox7 until the page was refreshed manually.

Involves refactoring parts of draftdesk service and dependents to use Subjects rather than trying to keep local copies of "froms" and "drafts" uptodate

Plus a couple of smaller commits for niggles: switch to empty messagelist when choosing empty folders (was failing without index), and catch some rest api errors (and report to console) when fetching message index and content updates).


gtandersen (approved):
Great work @castaway!

PR #1277, opened 2022-06-23
fix(maillist): Ensure we display errors when api returns them

### LINKED ISSUE #1275: Toggling read status is reverted (with local index)
**Describe the bug**
When marking one or more messages as read while using the local index, their state goes back to unread after a short while.

See also #537.

**To Reproduce**
1. Go to https://runbox.com/app and ensure the local index is enabled
2. Click to mark one or more messages as read
3. Wait a short while and they are reverted to unread state

**Expected behavior**
The read state should be maintained.

 - OS: macOS
 - Browser: Firefox

castaway:
See also backend changes to fix the actual problem

Fixes #1275

### DIFF
--- a/src/app/rmm/email.ts
+++ b/src/app/rmm/email.ts
@@ -31,7 +31,7 @@ export class Email {
           data => {
             const reply = data;
             if ( reply['status'] === 'error' ) {
-              this.app.show_error( reply['error'].join( '' ), 'Dismiss' );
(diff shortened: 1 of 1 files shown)


gtandersen (approved):

ilci3co:
The mark read/unread fail issue is still there.

castaway:
We haven't deployed the change to the production servers yet.

PR #1280, opened 2022-06-28
fix(folderlist): Check current folder counts for all folder types

### LINKED ISSUE #1182: Spam and Trash message counters are incorrect
**Describe the bug**
The Spam and Trash message counters sometimes show more than 0, but no messages are displayed in the list.

**To Reproduce**
1. Go to https://runbox.com/app
2. Observe the Spam and Trash message counts
3. Compare counts that are larger than 0 with the actual contents in the folders

**Expected behavior**
The Spam and Trash folders should always show the correct message counts.

See also #512.

castaway:
This enhances previous code that only checked the index counts against
the rest api counts. Now we also check non-indexed folders (Spam/Trash)
stored folder count values against actual counts.

Needs a rest api update

Fixes #1182

### DIFF
--- a/src/app/rmmapi/rbwebmail.ts
+++ b/src/app/rmmapi/rbwebmail.ts
@@ -67,6 +67,12 @@ export class FolderListEntry {
     }
 }
 
+export class FolderStatsEntry {
+    total: number;
+    total_unseen: number;
+    total_seen: number;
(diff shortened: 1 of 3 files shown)


gtandersen (approved):
Good call @castaway!

PR #1297, opened 2022-08-22
Castaway/704 check for updates immediately

### LINKED ISSUE #704: #Angular: App updates during login procedure and log in details are lost and have to be re-entered
Please stop the app from updating during the log in procedure when there is a new version. Log in details are lost when this happens and have to be re-entered :)

castaway:
Fixes #704 .. or improves anyway

### DIFF
--- a/src/app/updatealert/updatealert.service.ts
+++ b/src/app/updatealert/updatealert.service.ts
@@ -17,16 +17,35 @@
 // along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
 // ---------- END RUNBOX LICENSE ----------
 
-import { ApplicationRef, Injectable, NgZone } from '@angular/core';
+import { ApplicationRef, Injectable } from '@angular/core';
 import { SwUpdate, VersionReadyEvent } from '@angular/service-worker';
 import { MatLegacyDialog as MatDialog } from '@angular/material/legacy-dialog';
-import { UpdateAlertComponent } from './updatealert.component';
 import { environment } from '../../environments/environment';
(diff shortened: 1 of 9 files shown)


castaway:
Not entirely sure this works yet - isStable apparently only gets stable if no polling things are started/happening, so may need to check those too.

This can be improved by using activateUpdate instead of window.reload, and then subscribing to "unrecoverable state" to reload in case of issues, see https://v11.angular.io/guide/service-worker-communications

gtandersen (approved):
Great, @castaway!

src/app/updatealert/updatealert.service.ts
```diff
@@ -20,44 +20,68 @@
...
 
 @Injectable()
 export class UpdateAlertService {
+    public updateIsReady: BehaviorSubject<boolean> = new BehaviorSubject(false);
+    public updateStatus: UpdateStatus = {
+        'type':'UPDATE_AVAILABLE',
+        'current':
+        {'hash':'blah', 'appData':{'build_epoch':'XX'}},
+        'available':
+        {'hash':'blah', 'appData':{'commit':'test', 'build_time': 'time', 'build_epoch':'XX'}},
```
shadowbas, line 47: The blah seems a bit strange. What is the reason?

castaway: I added some default data, so that I could add a temporary button to simulate an update, so we could actually view/test the snackbar.. I'll remove it in case it actually ever gets accidentally used!

shadowbas (approved):

PR #1299, opened 2022-08-24
Checks the msg IDs given to API for certain endpoints are valid

piratefinn:
A frontend addressing of an issue in [internal issues] 1177 and 1178, verifying via a isInteger filter that the functions which call move_messages and and verifymessages for the searchindex that message IDs given are valid integers. Number.isInteger will return true on valid integer numbers (including 0) but false on empty/NaN values, ensuring at least they are valid integers before they are given to the API. The is because it doesn't otherwise check in the arrays (or array of objects) that the message ID's are valid.
This issue has also been addressed in the backend, but this is for consistency and cutting down redundant API calls that are found already invalid.
From a npm start and reporting existing spam mail in my SC mailbox it appears to function as frontend behaves and general usage doesn't report back network errors back from the backend.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -621,7 +621,9 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
   public trainSpam(params) {
     const msg = params.is_spam ? 'Reporting spam' : 'Reporting not spam';
     const snackBarRef = this.snackBar.open( msg );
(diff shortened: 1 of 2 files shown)


gtandersen (approved):
Thanks @piratefinn! Could you have a look as well @castaway? :)

castaway (approved):
Looks pretty sane to me

Issue #805, opened 2020-11-04
Calendar UI doesn't advertise the way to sync it between apps


gtandersen:
While RMM7 Contacts has a help function to explain how to sync it with other apps, Calendar does not.

At a minimum, we could add an obvious link to: https://help.runbox.com/using-a-calendar-client-with-caldav/

castaway:
Neither does webmail, to be fair (have to dig in help docs for smtp/imap links etc)

gtandersen:
@castaway That is true, and we should improve upon this as well -- see #816

gtandersen:
New issue #1304 created for the suggestion in #1301; closing this issue.

PR #1301, opened 2022-08-25
Adds a link to caldav sync guide on calendar

piratefinn:
Adds a link to the Caldav calendar syncing help as per #805 and appears as below in a new tab. Required some CSS tweaks as mat-list-item wasn't correctly taking certain custom styles in styles.css due to being used as an attribute to an anchor (as per angular docs specify it to be done for links).
Name of link is simply a first pass guess for "this is a guide on how to sync with Caldav" to avoid obscurity/confusion on the link purpose.
image

### DIFF
--- a/src/app/calendar-app/calendar-app.component.html
+++ b/src/app/calendar-app/calendar-app.component.html
@@ -129,6 +129,11 @@
                 <p mat-line>Settings</p>
             </mat-list-item>
 
+            <a mat-list-item target="_blank" href="https://help.runbox.com/using-a-calendar-client-with-caldav/" class="calendarMenuButton">
+                <mat-icon mat-list-icon svgIcon="sync"></mat-icon>
+                <p mat-line>CalDAV Sync Guide</p>
+            </a>
+
             <mat-divider></mat-divider>
 
             <div class="calendarsSideHeader">
--- a/src/styles.scss
+++ b/src/styles.scss
@@ -121,11 +121,16 @@ mat-form-field {
     font-size: 15px !important;
 }
 
-mat-list-item .mat-list-item-content {
+/* The anchor tag classing is required as it doesn't inherit the styles with html level a[mat-list-item] */
+mat-list-item .mat-list-item-content,
+a[mat-list-item] .mat-list-item-content {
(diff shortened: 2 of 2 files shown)


gtandersen (approved):
Excellent, @piratefinn!

piratefinn:
> Excellent, @piratefinn!

If the name is fine I'll merge it, though maybe @runboxdave has opinions on if it's a suitable name too.

runboxdave:
If we do this the same as we did in Contacts then the information for how to sync would be under the Settings and not have a separate menu entry. It makes sense to me to have it under settings but equally it could be a separate menu entry. In Contacts the section is headed "Your contacts client (CardDAV) settings " so we should probably us a similar name for Calendar.

I think if you asked me to choose where it should go I would say under Settings. You don't need to see it more than once in a a blue Moon so it's probably just an annoyance to have it as a main menu entry.

piratefinn:
> If we do this the same as we did in Contacts then the information for how to sync would be under the Settings and not have a separate menu entry. It makes sense to me to have it under settings but equally it could be a separate menu entry. In Contacts the section is headed "Your contacts client (CardDAV) settings " so we should probably us a similar name for Calendar.
> 
> I think if you asked me to choose where it should go I would say under Settings. You don't need to see it more than once in a a blue Moon so it's probably just an annoyance to have it as a main menu entry.

Considering the "Settings" for Calendar is a pop window instead and thus both lacks the screen real-estate of Contacts settings and is a whole page of info on that wiki link, I'm understanding here then that instead a "How to sync with Caldav" link should be on the settings as an obvious URL to that page is desired.

gtandersen:
While it's always useful to be consistent we are now discussing more than what the original issue asked for, and the layout for Calendar differs from Contacts where there's room for a settings pane to the right of the contacts.

Making these layouts be consistent could be decided in a separate issue (and potentially all settings could be moved to the general Settings area), and in the meantime I suggest we leave the link where it is since it's easy to find for users (and we could even add a similar link to the Contacts menu pane).

gtandersen:
New issue #1304 created for the above suggestion.

PR #1316, opened 2022-10-12
Separate out index/message updates to background worker

castaway:
Initiate a separate "web worker" thread to continually poll listallmessages and update the xapian index stored in indexddb. The stored index is a shared resource between the main and web worker threads, after each complete update the index is saved to indexddb and a message is sent to the main thread to refresh/reopen the xapian database. (This also updates the displayed message list if necessary)

If sync is turned off the separate worker thread still polls listallmessages, and informs the main thread which folder contents are "stale" (aka need refreshing).

The main thread manages fetching the actual message contents. Content updates are communicated back to the web worker thread to insert into the index (so search can find message content). Currently, as for existing code, this passes the entire text content of each message to the worker thread to store in the index.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -35,7 +35,8 @@
             "scripts": [
               "node_modules/tinymce/tinymce.min.js",
               "node_modules/tinymce/icons/default/icons.min.js"
-            ]
+            ],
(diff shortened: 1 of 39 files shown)


gtandersen (approved):
Excellent work @castaway!

PR #1321, opened 2022-10-19
Castaway/compose improvements

castaway:
Making an assumption that multiple drafts appearing is because of multiple save-runs happening at the same time, especially during the first save, which would send the "new draft" mid (a negative value) multiple times, and thus create multiple drafts. Fixed by only letting it happen once per draft.

Fixed a save issue where an initial save when in HTML mode wouldn't save until there was some content in the email body.

Clear attachment inputs after use (otherwise wouldnt trigger on attaching same file twice)


gtandersen (approved):
Excellent!

Issue #1269, opened 2022-06-07
Calendar: New all day events saved as timed ones


castaway:
**Describe the bug**
Creating a new "all day" event in runbox7 shows up correctly when creating, but after saving (and re-editing) shows up as a timed event instead of an all day one.

**To Reproduce**
1. Visit runbox.com/app/calendar
2. Click on '+' on a day for a new event
3. Enter a title, check the "all day" checkbox
4. Save event
5. Edit event - should reload as with "all day" checkbox checked and no times, but instead loads without checkbox checked, and with times

**Expected behavior**
Event is saved as an "all day" event (no times)

**Screenshots**
Creation: ![image](https://user-images.githubusercontent.com/109990/172408717-e660a6b9-e6e1-4e29-a704-a6c5669f1057.png)

Editing: 
![image](https://user-images.githubusercontent.com/109990/172408804-ac7d2832-0e2f-4677-b6a6-bcd031efd283.png)

**Desktop (please complete the following information):**
 - Linux
 - Vivaldi 
 - Version 5.2

PR #1325, opened 2022-10-25
fix(calendar): Setting "all day" flag on events now saves properly

### LINKED ISSUE #1292: Calendar events that are all day events are added across two days
**Describe the bug**
Calendar events that are annually recurring all day events are added across two days

**To Reproduce**
1. Add a new event
2. Set to all-day

Going back to the event the Details tab shows the all-day has  been turned off and times inserted instead.

**Expected behavior**
Event should be set correctly to the parameters set.

Might be related to https://github.com/runbox/runbox7/issues/745

castaway:
Fixes #1292 #1269


gtandersen (approved):
Excellent, @castaway!

PR #1326, opened 2022-10-26
Castaway/compose images

castaway:
1. Allow pasting, or drag&drop of images directly into HTML compose window
2. Insert/Edit dialog box lists attached images to insert into HTML compose window


gtandersen (approved):
Splendid!

PR #1343, opened 2022-12-13
Castaway/1122 drag and drop loses one

castaway:
When click & dragging: 
If message clicked on is already selected, do nothing and continue with drag
If message is not yet selected, select message and continue with drag

(Existing logic was somewhat convoluted)

Also added: We can now drag messages from Date, From/To, Subject and Size columns (instead of just Subject)


gtandersen (approved):
Great to have this fixed, @castaway!

PR #1368, opened 2023-01-16
Geir/overview improvements

gtandersen:


src/app/start/startdesk.component.ts
```diff
@@ -85,7 +86,7 @@ export class StartDeskComponent implements OnInit {
     // TODO: from appsettings or such?
     unreadOnly = true;
     timeSpan = TimeSpan.TODAY;
-    folder = FolderSelection.ALL;
+    folder = FolderSelection.INBOX;
```
castaway, line 89: FolderSelection.INBOX (defined on line 141) seems to only show "Inbox" which will ignore "INBOX" (which we still have an issue with, until dovecot imap import is sorted) .. is that ok?

gtandersen: @castaway Good point -- how does this look (tested and works)? https://github.com/runbox/runbox7/pull/1368/commits/50d52189bbd4978528159a8b7a75be55d2a99a3d

castaway: Yup that seems saner ;)

castaway (approved):

PR #1398, opened 2023-03-29
fix(compose): More readable attachment file size display

### LINKED ISSUE #627: Mail: attachment size value and units
When an attachment is being uploaded, its size is displayed in a human-readable number (e.g. 1.15 MB). This is perfect. Once uploaded, the size has no units and is not human readable (1207247). 

Ideally, the attachment size would have units and its value would be simple to read (so as in the first case).

castaway:
Fixes #627
2023-03-29 12 29 03 mallet3 scsys co uk 63a159a9d449


gtandersen (approved):
Excellent :)

PR #1401, opened 2023-03-31
style(security): Specify which special characters are allowed in passwords.

gtandersen:


castaway (approved):
The special character requirement is somewhat annoying (browser "suggest password" here doesnt contain one, nor has the option to force one).. however this is a suitable fix for the current issue

PR #1421, opened 2023-04-27
fix(compose): Generate Reply/Fwd header text only when needed

### LINKED ISSUE #234: "Undefined" ipo header when replying

castaway:
Fixes #234

For reasons best known to history, we were "generating" the reply/fwd headers (where header means "first line(s) of message" in the mail viewer template, then copying them out when doing a reply or fwd action. As pointed out in the #234 thread, quick actions could mean a message was opened, and the "reply" button clicked, before the reply header was generated, leading to "undefined" header strings.

I have moved the generation out of the template, and into the "create draft" code (which has the same mail object). Which means we could also now "reply" without opening a message, if we wanted to.

I also made the HTML+Text reply headers match, as one was regenerated differently!

### DIFF
--- a/src/app/compose/compose.component.ts
+++ b/src/app/compose/compose.component.ts
@@ -146,9 +146,9 @@ export class ComposeComponent implements AfterViewInit, OnDestroy, OnInit {
             if (this.composeInHTMLByDefault) {
                 this.model.useHTML = true;
             }
-            // This.. shouldnt happen as only have content if reply/fwd
+            // above was true and we replied to a text-view email:
             if (this.model.useHTML && !this.model.html) {
-                this.model.html = this.model.msg_body;
+                this.model.html = this.model.msg_body.replace(/\n/g, '<br />\n');
                 console.log('Copied msg body to html attribute');
             }
             if (this.model.cc.length > 0) {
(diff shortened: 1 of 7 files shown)


castaway:
2023-04-27 10 11 29 mallet3 scsys co uk c4ce72daf6a8
2023-04-27 10 10 59 mallet3 scsys co uk 5fc3e933e0cb
2023-04-27 10 10 43 mallet3 scsys co uk 32f0a932ea4e
2023-04-27 10 10 27 mallet3 scsys co uk 2572bcd00cec

runboxdave:
1. Replying without opening is a feature of Runbox 6 so might be worth introducing in Runbox 7.
2. I think adding the time zone when forwarding is useful.
3. I think adding the time zone when replying isn't needed and time should be local time to the user.

gtandersen:
> 1. Replying without opening is a feature of Runbox 6 so might be worth introducing in Runbox 7.
> 
> 2. I think adding the time zone when forwarding is useful.
> 
> 3. I think adding the time zone when replying isn't needed and time should be local time to the user.

Agree with these, although the time zone may also be included in replies for consistency.

castaway:
2023-05-02 15 50 30 mallet3 scsys co uk debd108b159b
2023-05-02 15 49 16 mallet3 scsys co uk 0ed397dcae16
2023-05-02 15 49 01 mallet3 scsys co uk 1916b6030333
2023-05-02 15 48 30 mallet3 scsys co uk ece243846964

Updated with Timezones

gtandersen (approved):
Looks good @castaway!

PR #1429, opened 2023-05-07
style(mailviewer): Increase vertical flexibility of subject field.

gtandersen:


castaway (approved):
looks good (got confused for a minute cos the code uses #messageContent .. but contents and content are not the same thing)

PR #1437, opened 2023-05-12
feat(search): Add more options to advanced search pane.

gtandersen:
Advanced search options


castaway:
This doesn't seem to quite work as expected, if I only check "Unread only" I get a blank list. I think its because "NOT X" isnt something that works in the default Xapian query parser: https://xapian.org/docs/queryparser.html

Possibly we should enable FLAG_PURE_NOT ?

PR #1441, opened 2023-05-22
feat(search): Enable Unread Only checkbox only if other options set

castaway:


gtandersen (approved):
Excellent, thank you :)

PR #1458, opened 2023-06-08
fix(messagecache): Change MessageCache to require a user ID

antoniskalou:
See https://github.com/runbox/private-issues/issues/94

### DIFF
--- a/src/app/rmmapi/messagecache.ts
+++ b/src/app/rmmapi/messagecache.ts
@@ -17,18 +17,16 @@
 // along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
 // ---------- END RUNBOX LICENSE ----------
 
-import { Injectable } from '@angular/core';
 import { Dexie } from 'dexie';
 import { MessageContents } from './rbwebmail';
 
-@Injectable()
 export class MessageCache {
     db: Dexie;
     message_version = 5;
 
-    constructor() {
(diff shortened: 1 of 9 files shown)


src/app/rmmapi/messagecache.ts
```diff
@@ -17,18 +17,16 @@
...
 
-@Injectable()
 export class MessageCache {
     db: Dexie;
     message_version = 5;
 
-    constructor() {
+    constructor(userId: number) {
         try {
-            this.db = new Dexie('messageCache');
```
castaway, line 31: Any thoughts on some sort of: if original "messageCache" indexdb exists, then delete (or rename!?) that before creating the new one using the userid? Otherwise we leave GBs of data lying around.

Generally we try and upgrade or at least throw out old storage we're not longer using. I suggest adding a comment which says "remove this chunk of code in 2024-06" or similar to it

castaway (commented):
Need some logic for deleting or converting existing messageCache indexdbs

src/app/rmmapi/messagecache.ts
antoniskalou: Good suggestion.

I'll add some code to delete the original `messageCache`. I'd rather not rename it because we're not 100% that its the correct message cache for the current user.

castaway (approved):
lgtm

Issue #1439, opened 2023-05-16
Message list action button blocked by tooltip popup of a previous button


kotecky:
**Describe the bug**
When selected, message list items have buttons (mark as read, mark as spam, ...). On hover on a button, a grey title popup appears underneath. In the case of the message lists, this blocks the view/access to the icon below and a user needs to move the mouse elsewhere to hide it. 

**To Reproduce**
1. Go to a message list, 
2. Select one, set of buttons appears, 
3. Hover to get the title displayed, 
4. It blocks the view/access to the button below. 

**Expected behavior**
The title should appear in a place where it does not block the view/access to the other buttons. Lest or right of the icon. In the case of a closed message pane in a smaller viewport the title would have been outside of the viewport if on the right. The left hand side then seems to be the only choice. 

**Desktop (please complete the following information):**
 - OS: all
 - Browser: any

gtandersen:
@kotecky Thanks for this report. The issue is confirmed and will be prioritized.

gtandersen:
This also affects other buttons such as the check-all checkbox in the message list.

PR #1466, opened 2023-07-05
fix(tooltips): Move context tool button tooltips to the left side

antoniskalou:
Currently tooltips show at the bottom of the button, blocking the button bellow it.

Because of some weird behaviour, the tooltip does not vanish when you move off it. This is probably due to an outdated material version, because when reviewing the official examples I do not experience the same behaviour
https://v6.material.angular.io/components/tooltip/overview

As a temporary fix, I have moved the tooltips to the left.

See https://github.com/runbox/runbox7/issues/1439

lefttooltip

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -314,11 +314,12 @@
   </div>
 
   <div class="contextToolButtons" *ngIf="!hasChildRouterOutlet && showSelectOperations && !showSelectMarkOpMenu">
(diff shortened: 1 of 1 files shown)


gtandersen:
Odd since this seems to have started happening recently. This fix should help though, until any underlying issue might be corrected in MD.

gtandersen (approved):

PR #1475, opened 2023-07-20
fix(account): Only subscription/renewal pages for expired accounts

### LINKED ISSUE #1473: Can't renew when account is expired
**Describe the bug**
Visiting account/subscriptions should always work when the account is expired

**To Reproduce**
1. Login
2. See "Expired, visit 'Your Subscriptions'" message
3. Click on "Your Subscriptions"
4. Get distracted and click on any other page, eg Mail icon in menu
5. Get redirected to the Login page as your account has expired
6. Login again, click on "Your Subscriptions"
7. This now sends you back to login, no chance to renew

**Expected behavior**
Visiting the subscriptions page should always work on an expired account

castaway:
Expired accounts should only be able to visit account/subscriptions (and related pages) to process a renewal. This prevents starting the index updates or current message list, as these will delete the session cookie when called with an expired account, which prevents the user from visiting the subscriptions page.

Fixes #1473

### DIFF
--- a/src/app/rmmapi/rbwebmail.ts
+++ b/src/app/rmmapi/rbwebmail.ts
@@ -92,6 +92,8 @@ export class RunboxMe {
     public is_trial: boolean;
     public uses_own_domain: boolean;
 
+    public account_status: string;
+
     public owner?: {
         uid: number;
         username: string;
@@ -114,11 +116,14 @@ export class RunboxMe {
     }
 
     getCreatedMoment(): moment.Moment {
-      return moment(this.user_created, 'X');
+        return moment(this.user_created, 'X');
     }
     newerThan(duration: number): boolean {
         const now = moment();
         return this.getCreatedMoment().diff(now) < duration;
+    }
+    isExpired(): boolean {
+        return this.account_status === 'expired';
     }
 }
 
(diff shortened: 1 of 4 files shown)


gtandersen (approved):

src/app/rmmapi/rbwebmail.ts
```diff
@@ -294,47 +296,55 @@ export class RunboxWebmailAPI {
...
-                    if (size === -1) {
-                        // Size = -1 means deleted flag is set - ref hack in Webmail.pm
-                        ret.deletedFlag = true;
-                    }
-                    return ret;
-                })
-            )
-        );
+          return this.me.pipe(
+            mergeMap((me) => {
```
antoniskalou, line 300: Seems fine to be, but couldn't you have done this as a filter?

```javascript
return this.me.pipe(
  filter(me => me.account_status !== 'expired'),
  mergeMap((me) => {
    // ...
  }
}
```

src/app/xapian/searchservice.ts
```diff
@@ -103,6 +103,7 @@ export class SearchService {
   localdir: string;
   partitionsdir: string;
 
+  isExpired: boolean = false;
```
antoniskalou, line 106: If you feel like it, it might be a good idea to add that as a method in `RunboxMe` instead since there's seems to be a lot of `me.account_status === 'expired'`

antoniskalou (commented):
Few minor things & some suggestions

src/app/rmmapi/rmmhttpinterceptor.service.ts
```diff
@@ -76,6 +77,19 @@ export class RMMHttpInterceptorService implements HttpInterceptor {
                             this.checkAccountStatus();
                         }
                     }
+                    if (req.url === '/rest/v1/me' && r.body) {
+                        this.currentMe = r.body.result;
```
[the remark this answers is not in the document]
castaway: Yes, am using .currentMe over different API calls, it should stick around for later use

antoniskalou (approved):

PR #1478, opened 2023-08-02
fix(accounts): Subscriptions links on account expiry

castaway:
Expired accounts should only be able to visit account/subscriptions, attempting to view anything else should return them to account/subscriptions

### DIFF
--- a/src/app/rmmapi/rmmauthguard.service.ts
+++ b/src/app/rmmapi/rmmauthguard.service.ts
@@ -19,7 +19,7 @@
 
 import { Injectable } from '@angular/core';
 import { HttpClient } from '@angular/common/http';
-import { CanActivate, ActivatedRouteSnapshot, RouterStateSnapshot, Router, CanActivateChild } from '@angular/router';
+import { CanActivate, ActivatedRouteSnapshot, RouterStateSnapshot, UrlTree, Router, CanActivateChild } from '@angular/router';
 import { Observable } from 'rxjs';
 import { map, take } from 'rxjs/operators';
 
@@ -28,6 +28,7 @@ export class RMMAuthGuardService implements CanActivate, CanActivateChild {
 
     urlBeforeLogin = '';
     wasLoggedIn = false;
+    currentMe;
 
     constructor(
         public http: HttpClient,
@@ -36,8 +37,8 @@ export class RMMAuthGuardService implements CanActivate, CanActivateChild {
 
     }
 
-    checkLogin(): Promise<boolean> {
-        return new Promise<boolean>((resolve, _reject) => {
+    checkLogin(): Promise<boolean | UrlTree> {
+        return new Promise<boolean | UrlTree>((resolve, _reject) => {
             this.isLoggedIn().pipe(take(1)).subscribe(
                 success => {
                     if (!success) {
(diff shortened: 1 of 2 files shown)


src/app/rmmapi/rmmauthguard.service.ts
```diff
@@ -58,18 +60,34 @@ export class RMMAuthGuardService implements CanActivate, CanActivateChild {
...
+                return this.checkStatus();
             }),
         );
     }
 
-    canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): boolean | Observable<boolean> | Promise<boolean> {
+    checkStatus(): boolean | UrlTree {
+        // Expired users are only allowed to visit account/subscriptions
+        if (this.currentMe && this.currentState
+            && this.currentMe.account_status === 'expired'
```
antoniskalou, line 78: Is it possible to do `this.currentMe.isExpired()` instead?

src/app/rmmapi/rmmauthguard.service.ts
```diff
@@ -58,18 +60,34 @@ export class RMMAuthGuardService implements CanActivate, CanActivateChild {
...
+        } else {
+            return this.wasLoggedIn;
+        }
+        
+    }
+
+    canActivate(route: ActivatedRouteSnapshot, state: RouterStateSnapshot): boolean | UrlTree | Observable<boolean | UrlTree> | Promise<boolean | UrlTree> {
         this.urlBeforeLogin = state.url;
 
+        this.currentState = state;
```
antoniskalou, line 90: Since you're only using the `.url` property of `RouterStateSnapshot`, wouldn't it be better to rename `currentState` to `currentURL` and change it to `this.currentURL = state.url`?

antoniskalou (approved):

src/app/rmmapi/rmmauthguard.service.ts
castaway: Nope, its not a RunboxMe obj, just a chunk of data (and it seems overkill to create the object to me)

src/app/rmmapi/rmmauthguard.service.ts
castaway: Probably, guess I did future proof that a tad..

antoniskalou: I like to keep the least amount of state around that I need, saved me some headaches. Its up to you though, PR approved

castaway: In fact your snippet just pointed out that the code is already saving the url .. so I've uselessly duplicated it! Yay for reviews

antoniskalou: Oh yea, lol, on the line right above it 😆

src/app/rmmapi/rmmauthguard.service.ts
```diff
@@ -100,10 +119,10 @@ export class RMMAuthGuardService implements CanActivate, CanActivateChild {
         this.router.navigate(['login']);
     }
 
-  redirectToSubscriptions() {
-      if (this.router.url !== 'account/subscriptions') {
-        console.log('Will navigate back to account/subscriptions');
-        this.router.navigate(['account/subscriptions']);
-      }
-    }
+    // redirectToSubscriptions() {
```
antoniskalou, line 122: Also, is this staying or should we remove it?

castaway: thats rubbish, will remove

Issue #644, opened 2020-06-18
#Drafts: Drafts are sometimes/often not deleted when the message is sent


runboxdave:
As in the subject, sometimes messages which are sent continue to persist as drafts.

kotecky:
I keep witnessing this. Best-guessing that it is occurring in conjunction with a third party client (Thunderbird and FairMail).

castaway:
Tested: create draft in runbox7, save as draft, then send using IMAP client - draft is still present in runbox7 after sending. Its gone from Drafts in the IMAP client.

gtandersen:
Reopening this as it is currently often happening in more than one account.

castaway:
Happening how/where? With IMAP / runbox7 interaction on same drafts of.. something else?

gtandersen:
I believe it occurs (inconsistently) when sending a message from Runbox 7 without interacting with any other interfaces.

@runboxdave Do we have any more details from users?

kotecky:
I have the issue but only go to the Drafts folder in the web app, seldom elsewhere. The combination of devices I use is Thunderbird on Win10/IMAP, FairMail on Android/IMAP, and the RB7 web app on Firefox on Win10 and Android.

At the very moment there are few drafts of sent items, visible in FairMail and in RB7.

castaway:
@kotecky thanks! Can I get a bit more detail? Do you create & send emails with all those clients? The drafts that are hanging around after you sent them, do you know which clients they were a) started with b) sent with?

gtandersen:
Confirming that drafts are intermittently left after sending with Runbox 7.

MoritzLost:
I'm also seeing this issue, it's very frequent. More often than not, a draft will remain in the drafts folder after sending the email.

gtandersen:
Thanks for the feedback -- we will increase the priority of this issue.

gtandersen:
This may seem to involve Runbox 6 as it can thus far only be reproduced in a test account while Runbox 6 is also open in the same browser.

MoritzLost:
@gtandersen In my case, this was definitely not the case – I have never used Runbox 6, and I have seen this issue with only a single Runbox 7 tab open.

kotecky:
I never use Rb6 but have the not deleted drafts despite. IMAP?

antoniskalou:
[...]

PR #1481, opened 2023-08-18
fix(drafts): Drafts are not deleted when the message is sent

antoniskalou:
See https://github.com/runbox/runbox7/issues/644#issuecomment-1651655948


castaway (approved):
Looks good to me!

PR #1496, opened 2023-09-20
feat(aliases): Add alias page

antoniskalou:


castaway (commented):
Looks like a good start/tidy up. Could we have some e2e and/or unit tests too?

castaway (approved):
Bit confused how the e2e tests pass (no mockserver.ts update?).

Can we use example.com instead of shadowcat in the tests?

runboxdave:
Comment upon reviewing this page:

1. The page takes along time to load (that could be because testing was on staging though)
2. Errors such as alias exists and when you try to add an alias you are not allowed to seem to have no nicely worded text and just seem to be in the form of the name of an error e.g. alias_exists
3. Text labels are very small.
4. Gaps between rows look about 3 times the height of a row itself meaning only a few aliases soon fill a page vertically. Aliases are a key feature of our service as we allow lots of them and lots of customers use them.
5. No instructions on what the page is for.

castaway:
Based on Daves comments, still some work to do.

I have also just noticed that the new alias creation does not add a corresponding/linked identity (as the RMM6 creation does), so we will also need some API/backend changes too

castaway:
Various buttons for add alias on the /account/identities and /account/ pages also need to link to the new aliases page

castaway:
https://github.com/shadow-dot-cat/runbox7/tree/rage/1350-alias-page - I created a branch with the template+css split out

gtandersen:
- [ ] Add instructions on what the page is for.
- [ ] Move "Create new alias" above the list of existing aliases.
- [ ] The page takes along time to load (that could be because testing was on staging though).
- [ ] Add fields for creating an alias directly on the page instead of via a pop-up.
- [ ] Convert the layout to a table with headings, which removes the need for labels and much spacing between rows.
- [ ] Add a column for Created timestamp.
- [ ] Make the fields editable directly in the table, but keep the Edit and Delete buttons. "Edit" can then make the fields editable and be changed to a "Save" button.
- [ ] Errors such as alias exists and when you try to add an alias you are not allowed to seem to have no nicely worded text and just seem to be in the form of the name of an error e.g. alias_exists.

Comment on standardization: While the Identities page is similar in that it lists items that could be displayed in a table, it includes more details (such as signatures) that might be difficult to fit into table cells. Therefore there's an argument for keeping it as it is while we generally standardize on a table format (we may also reconsider Identities at a later time).

PR #1503, opened 2023-10-12
fix(identities): Split out "username" profile and protect it

### LINKED ISSUE #1463: Do not allow users to delete last/only/main identity
**Describe the bug**
If the users main identity is deleted, Compose (and no doubt other things) does not work (no default "from")

**To Reproduce**
1. Go to 'Identities'
2. Click on the main identity
3. Select Delete
4. Attempt to Compose an email - compose window does not appear as there is no default "from" address (linked to identities)

**Expected behavior**
Identities page to refuse to remove the last/main identity, and compose to always create a new draft

 - OS: All
 - Browser all
 - Version n/a

**Additional context**
See also #771 (not sure how this one happens)

castaway:
Fixes #1463

2023-10-12 17 36 08 mallet3 scsys co uk cc33e531ada0

@gtandersen I'm passing you the baton for ideas  on how to actually make the "username" identity stand out, currently I just have it with a red highlight the same as the "Email not validated".. however I think it should probably be a solid border, or a different shade or something?


castaway:
New thought: Don't highlight the identity on the main page listing, instead, where the "delete" box would appear (which it doesnt cos you cant delete it), add an explanation?

src/app/profiles/profile.service.ts
```diff
@@ -0,0 +1,137 @@
...
+import { map } from 'rxjs/operators';
+import { BehaviorSubject, Observable } from 'rxjs';
+import { RunboxMe, RunboxWebmailAPI } from '../rmmapi/rbwebmail';
+
+export interface FromPriority {
+    from_priority: number;
+    id: number;
+}
+
+export class Identity {
```
antoniskalou, line 29: Is there a reason this is using snake case? Is it because its deserializing from the API?

src/app/profiles/profile.service.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    update(id, values): Observable<boolean> {
+        return this.rmmapi.updateProfile(id, values).pipe(
+          map((res: boolean) => {
+              this.refresh();
+              return res;
+          })
+        );
+    }
+    re_validate(id) {
```
antoniskalou, line 119: shouldn't `re_validate` be `reValidate`?

src/app/profiles/profiles.component.html
```diff
@@ -12,29 +12,32 @@ <h2>Default Identity</h2>
...
+                    address does. Below are identities you can use if you want to send messages from your aliases.
+                    These identities can be customised by adding a different From Name, Signature or Reply-to address.
+                    You can also change the Runbox domain an alias uses. You can't delete these identities as they are
                     tied to your aliases but you can manage your aliases by going to Alias Administration.</p>
-                <p style="color: crimson" *ngIf="rmm.alias?.aliases_counter?.current === rmm.alias?.aliases_counter?.total">
+                <p style="color: crimson" *ngIf="alias_limits?.current === alias_limits?.total">
                     You have reached the maximum allowed number of Runbox aliases.
                 </p>
-                <p>To manage your aliases visit <a target="rmm6" href="https://runbox.com/mail/account_alias">Alias Administration</a>.</p>
+                <p>To manage your aliases visit <a target="rmm6" href="/mail/account_alias">Alias Administration</a>.</p>
```
antoniskalou, line 40: This will be `/accounts/aliases` after my PR is merged.

src/app/profiles/profiles.editor.modal.html
```diff
@@ -98,33 +97,38 @@
...
-    <div class="warning-container" *ngIf="data.profile.reference_type == 'preference' && data.profile.reference.status === 1">
+    <div class="warning-container" *ngIf="identity.reference_type == 'preference' && identity.reference.status === 1">
         Email not validated. Check your email or
-        <a href="javascript:void(0)" (click)="resend_validate_email(data.profile.id)">
+        <a href="javascript:void(0)" (click)="resend_validate_email(identity.id)">
             re-send</a>.
     </div>
 
-    <p class="mat-small" style="text-align: center; margin-top: 0;" *ngIf="this.data.profile.type === 'aliases'">To manage aliases visit <a target="rmm6" href="https://runbox.com/mail/account_alias">Alias Management</a>.</p>
+    <p class="mat-small" style="text-align: center; margin-top: 0;" *ngIf="identity.type === 'aliases'">To manage aliases visit <a target="rmm6" href="https://runbox.com/mail/account_alias">Alias Management</a>.</p>
```
antoniskalou, line 114: Change to `href=/mail/account_alias`

src/app/profiles/profiles.editor.modal.ts
```diff
@@ -50,131 +48,101 @@ export class ProfilesEditorModalComponent {
...
+        @Inject(MAT_DIALOG_DATA) public identity: Identity
     ) {
         this.tinymce_plugin = new TinyMCEPlugin();
-        if (data && data.type) {
-            this.type = data.type;
-            delete data.type;
-        }
-        if (data.profile && data.profile.email) {
-            this.set_localpart(data);
+        if (identity.email) {
```
antoniskalou, line 60: This comes before the `if(!identity)` check, will this cause an issue with `if (undefined.email)`?

src/app/profiles/profiles.editor.modal.ts
```diff
@@ -50,131 +48,101 @@ export class ProfilesEditorModalComponent {
...
-        this.check_reply_to(this.data);
+        this.check_reply_to(this.identity);
     }
-    check_reply_to(data) {
-        if (data && data.profile.email && data.profile.reply_to &&
-            data.profile.reply_to !== data.profile.email) {
+    check_reply_to(identity) {
+        if (identity && identity.email && identity.reply_to &&
+            identity.reply_to !== identity.email) {
             this.is_different_reply_to = true;
```
antoniskalou, line 81: Preferably this would be a utility function, like `isDifferentReplyTo(identity)`, but I understand you might not want to change it since its how the whole component works.

src/app/profiles/profiles.editor.modal.ts
```diff
@@ -50,131 +48,101 @@ export class ProfilesEditorModalComponent {
...
-        req.subscribe(reply => {
-            if (reply['status'] === 'success') {
-                this.rmm.profile.load();
-                this.close();
-                return;
-            }
-            this.is_busy = false;
-        });
+        this.profileService.create(values).subscribe(
+            res => this.close
```
antoniskalou, line 119: `this.close()`, it won't call the function otherwise (I also write too much perl 😆)

src/app/profiles/profiles.lister.ts
```diff
@@ -30,25 +29,22 @@ import { MobileQueryService, ScreenSize } from '../mobile-query.service';
...
         mobileQuery: MobileQueryService,
     ) {
-        this.rmm.me.load();
         this.mobile = mobileQuery.screenSize === ScreenSize.Phone;
         mobileQuery.screenSizeChanged.subscribe(size => this.mobile = size === ScreenSize.Phone);
     }
 
     edit(item): void {
-        item = JSON.parse(JSON.stringify(item));
+        //item = JSON.parse(JSON.stringify(item));
```
antoniskalou, line 47: Delete it if we don't need it 🗑️

src/app/profiles/profiles.editor.modal.ts
```diff
@@ -274,31 +241,20 @@ export class ProfilesEditorModalComponent {
             setup: editor => {
                 self.editor = editor;
                 editor.on('Change', () => {
-                    self.data.profile.signature = editor.getContent();
+                    self.identity.signature = editor.getContent();
```
antoniskalou, line 244: I believe the new lambdas (fat arrows) preserve `this`, so it isn't necessary to alias `this` to `self`

antoniskalou: Arrow functions don't have their own [bindings](https://developer.mozilla.org/en-US/docs/Glossary/Binding) to [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this), [arguments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments), or [super](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super), and should not be used as [methods](https://developer.mozilla.org/en-US/docs/Glossary/Method).

Source: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Lexical_this

antoniskalou (changes requested):

src/app/profiles/profiles.component.html
antoniskalou: Ill update these myself after the PR gets merged

src/app/profiles/profiles.editor.modal.ts
castaway: ah another case of the codebase doing the same thing 2 different ways... compose doesnt bother with self. so have removed it here too..

antoniskalou (approved):

src/app/compose/draftdesk.component.ts
```diff
@@ -101,6 +101,8 @@ export class DraftDeskComponent implements OnInit {
                 }
             });
             this.draftModelsInView.splice(0, 0, ...newEntries);
+            this.draftModelsInView.sort((a,b) => a.mid < b.mid ? -1 : 1);
```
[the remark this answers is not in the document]
antoniskalou: Looked into it, should be fine

src/app/profiles/profiles.editor.modal.ts
```diff
@@ -33,148 +34,114 @@ import { TinyMCEPlugin } from '../rmm/plugin/tinymce.plugin';
...
-            delete data.type;
-        }
-        if (data.profile && data.profile.email) {
-            this.set_localpart(data);
-        }
-        if (!data || !Object.keys(data).length || !data.profile) {
-            data = { profile: {} };
+        if (!identity || !Object.keys(identity).length) {
+            identity = new Identity;
             const self = this;
```
antoniskalou, line 61: you don't need to create a `self` in this case, since the fat arrow `=>` retains `this` for you.

```javascript
identity.name = ['first_name', 'last_name'].map((attr) => {
  return this.profileService.me[attr];
}).join(' ');
```

antoniskalou (approved):

src/app/profiles/profiles.editor.modal.ts
castaway: That'll be a left over hernan-ism .. *fixed*

src/app/profiles/profile.service.ts
```diff
@@ -0,0 +1,143 @@
...
+              return res;
+          })
+        );
+    }
+    reValidate(id) {
+        this.rmmapi.resendValidationEmail(id).subscribe(
+            reply => {
+                this.refresh();
+                if ( !reply ) {
+                    // snackbar error msg?
```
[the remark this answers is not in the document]
castaway: actually the level above (rbwebmail) deals with the errors, so deleting this

Issue #1286, opened 2022-07-14
Alphabetical folder order should be case insensitive


gtandersen:
**Is your feature request related to a problem? Please describe.**
Folders starting with lower case letters are shown after folders starting with upper case letters, so that "bar" is shown after "Foo".

**Describe the solution you'd like**
Folders should be shown alphanumerically regardless of case, as in this example:

1. bar
2. Baz
3. Foo
4. qux

PR #1511, opened 2023-11-19
fix(folders): Alphabetical folder order should be case insensitive.

ff7762:
This PR is a fix for the order of folders in this issue https://github.com/runbox/runbox7/issues/1286

Below in the screenshot, you can see the current functionality of this issue, which matches the fix that was described in the ticket.
image

I have also written a test for this, to ensure maintainability of this functionality.

------------------------------------------------------

Questions about feature/bug bounties. 
(Haven't seen another channel for correspondence.) 

This is my first contribution as part of the feature bounty program. It wasn't market as suitable for starters but I wanted to give it a crack.

Is there a communications channel where payments are made for accepted solutions? 
I would like to start working on a few more of these issues - lots more to fix!
Also, does high priority = urgent in the payment tiers? I would expect that this solution fits the $100 tier.

Let me know if there are any issues, I will fix them asap.


src/app/changelog/changes.ts
castaway, line 1: Please don't submit the changelog in a PR, as we generate those on deploy

e2e/cypress/integration/folders.ts
castaway, line 1: The built-in sort() for the folder list is putting the inbox etc last, so you'll need to rethink the test

castaway (changes requested):
Looks like a good fix, just need to tweak the tests to be slightly more clever, thanks!

PR #1513, opened 2023-12-07
fix(messagecache): Load message contents whenever we encounter them

castaway:
There are many sentry-errors resulting from the message list drawing code attempting to fetch the message contents from the cache, when the cache hasnt been created or updated yet (that happens after the canvas has been painted with the message list)

This fills the cache as we look for the data (using promises so without interrupting the canvas drawing)

### DIFF
--- a/src/app/xapian/searchservice.ts
+++ b/src/app/xapian/searchservice.ts
@@ -811,15 +811,7 @@ export class SearchService {
           if (this.messageTextCache.has(rmmMessageId)) {
               this.currentDocData.textcontent = this.messageTextCache.get(rmmMessageId);
           }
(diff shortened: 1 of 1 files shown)


castaway:
Notes for reviewers:

The messageCache contains the actual message contents, generated / filled when getMessageContents is called. We don't need this data for the message list (unless summaries are on!), however we attempt to fill the message content slot every time we fetch the message data in general.

Before this update we were fetching from cachedMessageContents in getDocData, before the cache had been filled, so there was a lot of "this cache contains null". Instead I have made it call the full getMessageContents (aka updateMessageText), which is also what we call after the canvas has been painted

antoniskalou (approved):

PR #1522, opened 2024-01-08
fix(payment): Correct link to Sub-accounts.

gtandersen:


castaway (changes requested):
This could just be "#subaccounts" I think?

Issue #1046, opened 2021-07-08
Keep partially downloaded index if synchronization fails


gtandersen:
When synchronizing the message index the download may fail for various reasons such as network issues. In this case any fully downloaded index partitions should be retained in the browser so that the user may resume the rest of the download (instead of starting from scratch) and use the partial index in the meantime.

gtandersen:
This implies that there could be a "Resume index synchronization" button somewhere in the UI.

gtandersen:
This appears to happen less frequently now.

PR #1531, opened 2024-01-24
feat(install): Replaces index dialog prompt with sidebar buttons

### LINKED ISSUE #1509: Decrease size of initial index sync modal
The "Welcome to Runbox 7" modal with information about index synchronization is too dominating and now interferes with the "Welcome to Runbox" initial offer graphic (https://github.com/runbox/runbox7/pull/1498).

Welcome!

The modal could be moved to the lower left corner underneath the folder list and converted into a larger version of the existing "Synchronize index" button.

The text could then say:

"Runbox will synchronize with your device to give you an optimal webmail experience.
You can remove these data from your device later using the controls in this area. 
OK | Cancel"

castaway:
Fixes #1509

2024-01-24 16 40 35 mallet3 scsys co uk f1eab8f73271

"Download" button how starts the index download, and greys out the button. While downloading, the "Cancel" button can be used to cancel the process. (This can take a moment or two as its only interrupting the download between files). Once either finished downloading, or cancelled, the dialog goes away.. and the usual "Synchronize/Stop synchronizing" appears under the folder list.


gtandersen:
Excellent, this is a big improvement and especially if we can also make the index start downloading automatically when the page loads. That would allow the user to experience Runbox 7 optimally, which will be increasingly desirable with the growing number of up-to-date indices.

We then only need a "Cancel" button and if that is pressed it can be changed to "Delete" in order to allow the privacy concerned user to stop the download and/or delete the index. :)

Ideally there would also be a "resume" button, but that's another issue (https://github.com/runbox/runbox7/issues/1046).

castaway:
We could.. should we? At least on mobile devices folks might not want a bunch of data download?

gtandersen:
I think the app should be used to the fullest of its ability, and ideally/eventually download the index transparently in the background without disrupting the user. It's similar to IMAP in that regard, and even without the index will cache data locally anyway.

We could probably detect the device's data connection type and prompt the user if it's not on wifi.

castaway:
image

Suggested location from @gtandersen

antoniskalou (approved):

PR #1546, opened 2024-03-25
style(payment): Update Coinbase payment modal.

gtandersen:
Ref https://help.coinbase.com/en/migrating-to-the-onchain-payment-protocol and https://github.com/runbox/runbox7/issues/1549.

Coinbase UI new


castaway (approved):
lgtm

castaway:
Oops, re-reviewed after Dave asked about non-coinbase .. This update removes the use of mainOptions from the html, are these no longer used / appropriate? if so we should removed all other references to them

gtandersen:
> Oops, re-reviewed after Dave asked about non-coinbase .. This update removes the use of mainOptions from the html, are these no longer used / appropriate? if so we should removed all other references to them

I will re-review this as well because we'll apparently need a way for customers to pay manually to Coinbase (without the Coinbase Commerce interface).

gtandersen:
In order to satisfy customers' desire to without using Coinbase, the additional button below is intended to trigger `initiateCryptoPayment()` similar to the `initiatePayment()`  in `shopping-cart.component.ts` and take the user to a cryptocurrency version of the payment receipt page with instructions on how to proceed:

Coinbase Transfer

gtandersen:
Updated payment instructions with addresses and QR codes for common currencies:

Cryptocurrency transfer

Cryptocurrency transfer 2

gtandersen:
We should rename "bitpay" to "coinbase" for cryptocurrency payments.

PR #1548, opened 2024-04-04
style(identities): Improve indication of Account Username.

gtandersen:
Third attempt :)

Account Username
Edit Identity Account Username


castaway:
I think this looks a bit odd in the "Email" section, can we just add it to the top of the entire card instead? @runboxdave thoughts? (see first image)

runboxdave:
Yes, it would possibly look better at the top of the entire card as you say.

gtandersen:
Something like this perhaps?

Identities Account Username

PR #1576, opened 2024-05-28
Update TinyMCE to v6, enable drag&drop of images into compose

castaway:
Drag&drop images uploads the image as an attachment, then updates the img src attribute to load the image using the attachment url. Sending then replaces it with a cid: id linking to the attached image.


gtandersen (approved):
Excellent!

PR #1580, opened 2024-06-17
fix(identities): Re-enable editing of preferred domain for aliases

castaway:
Broken by recent identities changes, alias identities should be allowed to change their "preferred domain", eg from runbox.com to runbox.uk


gtandersen (approved):

runboxdave:
Tested and works.

Issue #1555, opened 2024-04-16
#Mail: Deleting a large number of messages fails with HTTP 400 or 414


gtandersen:
**Describe the bug**
When deleting a large number of messages, the server responds to the DELETE request with:

- `400 Bad request / Your browser sent an invalid request.` by the proxy server when the local index is not in use.
- `414 Request-URI Too Large` by nginx when the local index is in use.

**To Reproduce**
1. Go to https://runbox.com/app
2. Select a large number (hundreds) of messages.
3. Click to delete the messages.
4. Observe that the messages disappear for a few seconds and then return.

**Expected behavior**
All the selected messages should be deleted (moved to Trash).

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser : Firefox

**Additional context**
Moving a large number of messages with or without the local index works as expected, presumably because the request is a POST and not a DELETE. However, if moving messages without the local index they will still appear in the source folder on other devices that use the local index, possibly because moving messages to Trash (as opposed to deleting) does not include them in https://runbox.com/rest/v1/list/deleted_messages/.

See also https://github.com/runbox/runbox7/issues/1244

castaway:
rest api / nginx settings: https://medium.com/aviabird/413-414-request-url-entity-too-large-error-nginx-b6dcece6f5dd
* client_max_body_size set to 130m
* large_client_header_buffers unset, default is 8 KB

PR #1595, opened 2024-09-05
fix(messages): Use batch_delete for deleting multiple messages

antoniskalou:
See https://github.com/runbox/runbox7/issues/1555


castaway (approved):
lgtm

PR #1606, opened 2024-10-16
fix(config): Opt-out of angular analytics

bas080:
Set the angular analytics to be disabled using the angular config.


castaway (approved):
effectively disabled by default anyway.. might as well be explicit

PR #1613, opened 2024-10-24
feat(templates): Create draft on template item click

shadowbas:


src/app/compose/draftdesk.service.ts
```diff
@@ -309,6 +309,27 @@ export class DraftDeskService {
         this.draftModels.next(models);
     }
 
+    public async newTemplateDraft(
+        messageId: number,
+    ) {
+        forkJoin([
+            this.rmmapi.getMessageFields(messageId),
```
castaway, line 316: You discovered a thing I didn't know existed..

I don't think we need getMessageFields here, as despite its name, getMessageContents actually stores the entire message object, see fetchMessageJSON in mailviewer/singlemailviewer.component.ts

castaway (changes requested):
minor changes required

We can probably remove getMessageFields to make things less confusing (it doesnt seem to be used anywhere)

src/app/compose/draftdesk.service.ts
shadowbas: 2024-10-29-171952_611x250_scrot

getMessageContents does not contain any fields. It is either the HTML or text of the message.

shadowbas: Oo I see, the headers have that information. That is unusual to me.

castaway: ah yeah the class doesnt contain alllll the fields (hysterical raisins no doubt)

src/app/compose/compose.component.html
```diff
@@ -24,13 +24,16 @@
             </button>
             <button *ngIf="!editing" mat-icon-button (click)="editDraft()" matTooltip="Edit draft" id="editDraftIcon">
                 <mat-icon svgIcon="pencil"></mat-icon>
-            </button>                 
+            </button>
+            <button *ngIf="editing" mat-icon-button matTooltip="Save as template" (click)="saveTemplate(true)" id="sendMail">
```
castaway, line 28: The "id" needs changing here - should be unique

src/app/compose/draftdesk.service.ts
```diff
@@ -240,6 +240,7 @@ export class DraftDeskService {
 
     constructor(public rmmapi: RunboxWebmailAPI,
                 private messagelistservice: MessageListService,
+                private rbwebmailapi: RunboxWebmailAPI,
```
castaway, line 243: Its already got a  RunboxWebmailAPI, as "mmapi"

castaway (changes requested):

src/app/compose/compose.component.html
shadowbas: Good catch, is an id even necessary? I'll have a look.

shadowbas: I guess we do need them for the automated tests

castaway:
Looks fairly sane to me - one minor thought - can we edit/save existing Templates without creating new ones? (I think thats what the tid/mid are for.. tid=existing template id, mid=message based on template)

castaway (approved):
lgtm!

PR #1616, opened 2024-11-04
feat(a11y): Remove underline on anchor hover

shadowbas:
Make default link color the same as link theme color and not browser default.


castaway (approved):
good idea

Issue #1617, opened 2024-11-11
#Install: Update app on new release, app freezing just displaying runbox logo


castaway:
**Describe the bug**
Sometimes the runbox7 application freezes, just displaying the Runbox Logo mid screen and nothing else

**To Reproduce**
1. Keep an open tab with runbox7
2. Wait .. return to it after an application update
3. Refresh the browser, may only have the logo on white screen

**Expected behavior**
App to display the dialog which prompts the user to update the application

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Linux
 - Browser Vivaldi
 - Version 7

**Additional context**

If we stop the code in devtools:
ngsw-worker stops with an uncaught error in cacheBustedFetchFromNetwork, `throw new SwCriticalError(`Hash mismatch (cacheBustedFetchFromNetwork): ${req.url}: expected ${canonicalHash}, got ${cacheBustedHash} (after cache busting)`);` Which appears to mean that the calculated hash of the fetched code does not match the hash in nsgw.json

castaway:
Can we (at least initially) catch the thrown error (in updatealerts.service.ts?) and force a refresh?

PR #1618, opened 2024-11-13
feat(worker): Add sentry to xapian worker scope

shadowbas:
Add sentry to the Xapian worker.

Related: #1617


src/build/add-sentry.js
shadowbas, line 1: @castaway , this approach of code gen and using git to revert is very messy and was causing me issues. I suggest we have this one env file that is generated using the `./gen-env.js` script. We use this module instead of interpolating writing and reverting with git.

.eslintrc.json
```diff
@@ -50,6 +50,20 @@
       "rules": {
         "@angular-eslint/template/eqeqeq": "warn"
       }
+    },
+    {
+      "files": [
+        "*.js"
```
shadowbas, line 56: Consider only applying this to the build directory.

src/app/sentry.ts
```diff
@@ -17,4 +17,9 @@
 // along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
 // ---------- END RUNBOX LICENSE ----------
 
-// ADD SENTRY HERE
+import { environment } from '../environments/environment'
+import * as Sentry from '@sentry/browser';
+
+Sentry.init({
+  dsn: environment.SENTRY_DSN
```
shadowbas, line 24: Could add to capture console.error messages:

      integrations: [Sentry.captureConsoleIntegration()],

shadowbas: Got this working after an update of typescript, angular and sentry. Integrations are not supported by the typescript version we are using. Hence the typescript and angular update.

castaway (approved):
Looks pretty good and seems to run .. only issue I have is with the access-settings (first e2e test) failing locally

PR #1627, opened 2024-11-19
build(app): Update build ngsw.json with appData

shadowbas:
Set the appData after re-calculation of ngsw file hashes as it overwrites it on re-calculation.

### DIFF
--- a/src/build/post-build.js
+++ b/src/build/post-build.js
@@ -1,10 +1,28 @@
-const execSync = require('child_process').execSync;
+const path = require('node:path');
+const execSync = require('node:child_process').execSync;
 const chalk = require('chalk');
+const fs = require('node:fs')
 
 execSync('git checkout src/app/buildtimestamp.ts');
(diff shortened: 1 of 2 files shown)


src/build/post-build.js
```diff
@@ -1,11 +1,29 @@
...
 
 execSync('git checkout src/app/buildtimestamp.ts');
 execSync('git checkout ngsw-config.json');
 execSync('npx sentry-cli sourcemaps inject ./dist/');
 execSync('npx ngsw-config ./dist ngsw-config.json "/app/"');
 
+const ngswPath = path.join(__dirname, '../../dist/ngsw.json')
+const ngswJSON = require(ngswPath);
+
+const build_time = new Date().toJSON();
```
shadowbas, line 14: In the pre-build.js we also create a build_time. I don't believe it to be an issue that they differ. Agred @castaway?

castaway (approved):
lgtm

PR #1634, opened 2024-11-29
feat(a11y-table): Define a standalone table component

shadowbas:
Add table to the app component.

- [x] Render massive amounts of data (tested with a million records)
- [x] Advanced (range) select features
- [x] Keyboard friendly with Tab and Enter.
- [x] Toggle preview with UI button
- [x] Use arrow keys to move up and down a message.
- [x] ~~Use j and k to scroll up and down in messages.~~ Dropped the support for j and k scroll.
- [x] Allow passing templates to render the cells (allows for more rich components).
- [x] Make the canvastable inert and not actually render to screen.
- [x] Drag and drop to folders.
- [x] Bulk actions on selected items.
- [x] Select row when navigating on fragment.
- [x] Show to in sent folder instead of from
- [x] Remove canvastable code as much as possible.
- [x] Add the flag icon on flagged messages.
- [x] Add aria-label to attachment icon.
- [x] Add aria-label to flagged messages.
- [x] Fix issue where ctrl+tab causes ctrl flag to stay true.
- [x] Better styling for narrow screen (break out of the table layout by using different display, considering flex)
- [x] Should only get the doc data when the item is in view (causing massive amount of requests)
- [x] Resize-able columns.
- [x] Fix the width of columns on initial render. This prevents jumping of columns during scroll.
- [x] Set sane defaults for initial columns widths.
- [x] Sort on column heading click.
- [x] Fix to table when index-sync-off in sent folder
- [x] Double click on column resize separator resets width
- [x] Figure out how to reset column widths on switching routes
- [x] Fix bug where last selected item is removed when no longer in view (selects first item)
- [x] Could we add a more visible/intuitive column divider indicator in the column header?
- [x] There doesn't seem to be a folder column in search results, which would be useful to have.
- [x] Fix issue where new rows causes weird scroll jump upward
- [x] Add answered column
- [x] Fix issue where columns are very narrow on initial load
- [x] Reduce font-size in the message list
- [x] Fix issue where bulk action buttons are no longer visible
- [x] Remove Direction.None from sort button cycle
- [x] Fix threaded view
- [x] Fix tests to work with html table
- [x] Remove console. calls
- [x] Autosquash and force push
- [x] Would be nice to have a sticky table head
- [x] Remove commented out code
- [x] Only perform column width reset when resize is vertical
- [ ] Re add updating index https://github.com/runbox/runbox7/pull/1634/files/bf4ca21fcaea8660712ec9615b82a3da644b9eee#r2163563214
- [x] Remove the `$` postfix for rxjs stream variables


src/app/accessible-table/accessible-table.component.ts
```diff
@@ -0,0 +1,119 @@
...
+    private elementRef: ElementRef,
+  ) {
+    // for (let i = 0; i < 1000; i++) {
+    //   this.emails.push({
+    //     date: new Date(2024, 10, i % 30 + 1).toISOString(),
+    //     from: `user${i + 1}@example.com`,
+    //     subject: `Subject ${i + 1}`,
+    //     size: Math.floor(Math.random() * 20) + 5
+    //   });
+    // }
```
shadowbas, line 33: Remove, was only used for testing.

src/app/accessible-table/accessible-table.component.ts
```diff
@@ -0,0 +1,221 @@
...
+
+    const selected = isSelected
+      ? this.selectedSet()
+      : new Set([row]);
+
+    this.selectionDragStarted.emit({
+      event,
+      selected,
+    })
+  }
```
shadowbas, line 219: I would have liked to decouple this from the table. It would be nice to have a better abstraction to work with draggable.

Maybe the complete table API is strange. Since when are tables concerned about selectedRows? Might it not be more correct to decouple that too by offering a better templating API. I have ideas but I also need to draw a line and say it's good enough.

src/app/app.component.scss
```diff
@@ -68,3 +72,10 @@
     width: 150px;
 }
 
+.text-primary {
+    color: #01579b;
```
shadowbas, line 76: Move to a css property

src/app/app.component.ts
```diff
@@ -1419,6 +1428,57 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
...
     }
   }
-}
 
-// vim: set shiftwidth=2
+  get selectedMessageIds() {
+    return Array.from(this.selectedRows).map(x => x.id);
+  }
+
+  updateRows() {
```
shadowbas, line 1436: This is where I normalize the sync messages and non-sync messages. It's a mess but the simplest I could thing of

src/app/app.component.ts
```diff
@@ -1419,6 +1428,57 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
...
+
+    this.rows = (() => {
+      if (!this.canvastable?.rows?.rows?.length) return []
+
+      if (Array.isArray(this.canvastable.rows.rows[0]))
```
shadowbas, line 1441: This is for when the rows are the arrays where the first item is the messageid.

gtandersen (approved):

src/app/app.component.ts
```diff
@@ -1419,6 +1474,179 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
...
+  updateRows() {
+    this.rows = this.canvastable?.rows?.rows ?? []
+
+    return this.enrichRows()
+  }
+
+  async enrichRows() {
+    const { start, end } = this.renderedRange
+
+    this.rows = await Promise.all(mapOverIndexes(async (value) => {
```
shadowbas, line 1493: This might cause a race condition. When renderedRange changes twice; it might happen that the first range change resolves later than the second range change.

- [ ] Find a way to merge resolved arrays in a way that will keep the latest and not overwrite fetched rows

src/app/accessible-table/accessible-table.component.html
```diff
@@ -0,0 +1,13 @@
+<cdk-virtual-scroll-viewport [itemSize]="firstRowHeight" class="email-viewport">
```
shadowbas, line 1: - [ ] Rename to virtual-scroll-table.component

src/app/accessible-table/accessible-table.component.ts
```diff
@@ -0,0 +1,110 @@
...
+} from '@angular/core';
+import { MatCheckboxModule } from '@angular/material/checkbox';
+import { CommonModule } from '@angular/common';
+import { ScrollingModule, CdkVirtualScrollViewport } from '@angular/cdk/scrolling';
+import { ListRange } from '@angular/cdk/collections';
+import { Subscription } from 'rxjs';
+import { debounceTime } from 'rxjs/operators';
+
+@Component({
+  selector: 'app-accessible-table',
```
shadowbas, line 43: - [ ] Rename to `app-virtual-scroll-table`

src/app/app.component.html
```diff
@@ -399,16 +399,229 @@ <h3 class="noMessageSelectedNotification">No Message Selected</h3>
...
+                      focussed: lastCheckedIndex === index,
+                      selected: rowSelectionModel.isSelected(item),
+                      checked: rowsSelectionModel.isSelected(item)
+                    }"
+                  >
+                    <tr>
+                      <td class="checkbox-cell">
+                        <mat-checkbox
+                          (click)="onCheckboxClick($event, item, index)"
+                          [checked]="rowsSelectionModel.isSelected(item)"
```
shadowbas, line 503: - [ ] Remove event listener from the skeleton version of a row.

src/app/app.component.scss
```diff
@@ -2,6 +2,10 @@
     border-right: 1px solid darkgrey !important;
 }
 
+canvastablecontainer {
+    opacity: 0;
+}
+
```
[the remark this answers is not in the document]
shadowbas: - [ ] Remove this.canv and the canvas element from canvastable

src/app/app.component.scss
```diff
@@ -68,3 +72,125 @@
...
+            &.subject {
+                width: 100%;
+            }
+        }
+    }
+
+}
+
+.skeleton-bone {
```
shadowbas, line 173: Move skeleton styles to global styles.

src/app/human-bytes.pipe.ts
```diff
@@ -0,0 +1,39 @@
...
+      return '0 B';
+    }
+
+    const base = 1000;
+    const suffixes = ['B', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+    const exponent = Math.floor(Math.log(value) / Math.log(base));
+
+    const result = (value / Math.pow(base, exponent)).toFixed(decimalPlaces);
+    return `${parseFloat(result)} ${suffixes[exponent]}`;
+  }
```
shadowbas, line 38: It appears we have a function in `src/app/messagetable/messagetablerow.ts`. The converting of a number that represents the amount of bytes to a human readable format is not specific to messagetable.

-  [ ] Make a human-readable utility module and use that here and in the messagetablerow.ts.

src/app/models/bindable-selection-model.ts
```diff
@@ -0,0 +1,44 @@
...
+  get selected(): T | T[] {
+    return this.selectionModel.isMultipleSelection() ? this.selectionModel.selected : this.selectionModel.selected[0];
+  }
+
+  // Setter for `selected`
+  set selected(items: T | T[]) {
+    const selection = (this.selectionModel.isMultipleSelection() ? items : [items]) as T[];
+    this.selectionModel.setSelection(...selection)
+  }
+}
```
shadowbas, line 44: Angular allows for two way binding in the template. Here we wrap the angular Selection model and add the selected property to be used for binding.

src/app/models/filter-selection-model.ts
```diff
@@ -0,0 +1,38 @@
...
+          return (...items: T[]) => {
+            return target.select(...items.filter(predicate));
+          };
+        }
+
+        return target[prop];
+      }
+    });
+  }
+}
```
shadowbas, line 38: A SelectionModel that takes a predicate which is used to filter items selected using the .select(...items) method. This is handy to prevent selecting items that have not been loaded yet. This is relevant when doing a range select which contains items that have not been loaded into array yet.

src/app/resizable-button/resizable-button.component.ts
```diff
@@ -0,0 +1,142 @@
...
+    parentElement.style.width = this.initialWidth;
+    setTimeout(() => {
+      this.setAbsoluteWidth()
+    }, 10)
+  }
+
+  @HostListener('window:resize')
+  onWindowResize() {
+    this.resetWidth();
+  }
```
shadowbas, line 72: Should this be here? When do we want to reset the widths?

src/app/websocketsearch/websocketsearchmaillist.ts
```diff
@@ -98,4 +98,15 @@ export class WebSocketSearchMailList extends MessageDisplay {
...
+      return {
+        id: this.getRowMessageId(rowIndex),
+        selectbox: this.isSelectedRow(rowIndex),
+        messageDate: this.getRow(rowIndex).dateTime,
+        from: this.getRow(rowIndex).fromName,
+        subject: this.getRow(rowIndex).subject,
+        size: this.getRow(rowIndex).size,
+      };
+    }
+
```
shadowbas, line 111: @castaway , could you please help with how to test the websocketsearchmaillist?

src/app/xapian/searchmessagedisplay.ts
```diff
@@ -42,11 +42,11 @@ export class SearchMessageDisplay extends MessageDisplay {
   getRowMessageId(index: number): number {
     let msgId = 0;
     try {
-      msgId = this.searchService.getMessageIdFromDocId(this.rows[index][0]);
+      msgId = this.searchService.getMessageIdFromDocId(this.getRowId(index));
     } catch (e) {
       // This shouldnt happen, it means something changed the stored
       // data without updating the messagedisplay rows.
-      console.log('Tried to lookup ' + index + ' in searchIndex, isnt there! ' + e);
+      console.error('Tried to lookup ' + index + ' in searchIndex, isnt there! ' + e);
```
shadowbas, line 49: - [ ] Change `+ e` to `, e` for more useful log in console

src/app/xapian/searchmessagedisplay.ts
```diff
@@ -215,4 +215,37 @@ export class SearchMessageDisplay extends MessageDisplay {
...
+      plaintext: this.searchService.getDocData(this.getRowId(index)).textcontent?.trim(),
+      size: this.searchService.api.getNumericValue(this.getRowId(index), 3),
+      attachment: this.searchService.getDocData(this.getRowId(index)).attachment ? true : false,
+      answered: this.searchService.getDocData(this.getRowId(index)).answered ? true : false,
+      flagged: this.searchService.getDocData(this.getRowId(index)).flagged ? true : false,
+      folder: this.searchService.getDocData(this.getRowId(index)).folder,
+      seen: this.searchService.getDocData(this.getRowId(index)).seen,
+    };
+
+    if (app.viewmode === 'conversations') {
```
shadowbas, line 236: - [ ] Add the conversation count also to the accessible table as it was in canvastable.

src/app/xapian/searchservice.ts
```diff
@@ -258,10 +258,10 @@ export class SearchService {
       FS.syncfs(true, () => {
           // console.log('Main: Syncd files:');
           // console.log(FS.stat(XAPIAN_GLASS_WR));
-          FS.readdir(this.partitionsdir).forEach((f) => {
+          // FS.readdir(this.partitionsdir).forEach((f) => {
             // console.log(`${f}`);
             // console.log(FS.stat(`${this.partitionsdir}/${f}`));
-          });
+          // });
```
shadowbas, line 264: Removed the look as it does nothing but block the thread. Easy performance win

castaway:
Notes while testing/poking around:
2025-02-04 11 50 40 rmmstage02 runbox com 1ff05c40c6c9
Clicking on the Arrow seems to do nothing other than make the arrow disappear (does not redisplay sorted date in reverse)

castaway:
Starting new comment since that one seems full:
11. Select-drag over checkboxes with mouse (works on prod) - doesnt here
12. No way of "select all on page" / "select all in folder" that I can see?

castaway:
Another odd issue: If I navigate away from the Inbox/message list (eg to Settings), then back via clicking on the menus at the top of the app - the message list shows up empty apart from the checkboxes.. If I swap browser tabs (eg to mattermost and back) then the list reappears ... 

2025-02-05 11 11 40 rmmstage02 runbox com 6718e7cce586

castaway:
Some testing this afternoon:
* The fetch of message contents according to the current message list - this should happen in both the index-sync and index-off modes
* If I click "Stop Index Synchronisation" things go a bit screwy!
2025-03-19-162146_1096x677_scrot

shadowbas:
> Some testing this afternoon:
> 
> * The fetch of message contents according to the current message list - this should happen in both the index-sync and index-off modes
> * If I click "Stop Index Synchronisation" things go a bit screwy! (hmm not always tho.. yay)
>   2025-03-19-162146_1096x677_scrot

@castaway , I am unable to reproduce this on my account. I could do a null check and drop the message when it is null.

gtandersen:
Improved default column widths:
Message table

castaway (changes requested):
Some queries/changes

src/app/xapian/searchmessagedisplay.ts
```diff
@@ -32,7 +32,7 @@ export class SearchMessageDisplay extends MessageDisplay {
   }
 
   getRowSeen(index: number): boolean {
-    return this.searchService.getDocData(this.rows[index][0]).seen ? false : true;
+    return this.searchService.getDocData(this.getRowId(index)).seen;
```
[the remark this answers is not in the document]
shadowbas: I'll just revert it to what it was

castaway (changes requested):

src/app/app.component.ts
```diff
@@ -470,35 +513,12 @@ export class AppComponent implements OnInit, AfterViewInit, CanvasTableSelectLis
...
-      this.rmmapi.downloadMessages(messageIds).then(
-        (messages) => {
-          const updateWorker = new Map();
-          for (const msg of messages) {
-            this.searchService.updateMessageText(msg['mid']);
-            updateWorker.set(msg['mid'], msg.text.text);
-          };
-          // Send to the messageCache in the worker, so we can add the text to the index:
-          if(updateWorker.size > 0) {
-            this.searchService.indexWorker.postMessage({'action': PostMessageAction.messageCache, 'updates': updateWorker });
```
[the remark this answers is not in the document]
shadowbas: Not sure what I broke by removing this. I'll add it to the todo of this PR.

castaway:
Copy this here so we don't lose track of it:

Testing on Thurs Oct 30th (screenshots in chat)

* selected some messages in the Inbox list
* marked them as Spam
* checked that the backend confirmed that they should now be in the Spam folder
* some of them disappeared from the Inbox message list and .. some did not
* also it only fetched the contents of the message after I clicked on it to view it (which was from the inbox, after it should have been moved to Spam)

PR #1643, opened 2024-12-19
Updates bulk messages PR, adding index working (less data passing) improvements

castaway:
On canvas update, fetch all in-view message contents at once (instead of one request per message), .. if not already in the cache. After fetching, update the index (to add the message contents to the index), for all the fetched messages at once, making less data exchanges between the two threads.

This should mean less busy threads and more prompt actions for users

### DIFF
--- a/e2e/mockserver/mockserver.ts
+++ b/e2e/mockserver/mockserver.ts
@@ -246,39 +246,26 @@ END:VCALENDAR
                 }
 
             }
+
+            const bulkemailendpiont = requesturl.match(/\/rest\/v1\/email\/download\/([0-9,]+)/);
+            if (bulkemailendpiont) {
+                const ids = bulkemailendpiont[1].split(',').map(id => parseInt(id, 10));
+
+                const messages = {};
+                for (const id of ids) {
+                    messages[id] = { json: this.getMessage(id).result };
+                }
+
+                response.end(JSON.stringify({
+                    'status': 'success',
+                    'result': messages,
+                }));
+            }
+
             const emailendpoint = requesturl.match(/\/rest\/v1\/email\/([0-9]+)/);
             if (emailendpoint) {
-                const mailid = emailendpoint[1];
-                let message_obj = mail_message_obj;
-                if (mailid === '11') {
-                    message_obj = JSON.parse(JSON.stringify(mail_message_obj));
-                    const to = message_obj.result.headers['to'];
-                    delete message_obj.result.headers['to'];
-                    message_obj.result.headers['cc'] = to;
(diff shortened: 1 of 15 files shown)


e2e/mockserver/mockserver.ts
```diff
@@ -246,39 +246,26 @@ END:VCALENDAR
                 }
 
             }
+
+            const bulkemailendpiont = requesturl.match(/\/rest\/v1\/email\/download\/([0-9,]+)/);
```
shadowbas, line 250: Fix typo and casing from  `bulkemailendpiont` to `bulkEmailEndpoint`?

src/app/rmmapi/rbwebmail.ts
```diff
@@ -265,6 +273,76 @@ export class RunboxWebmailAPI {
...
+                            reject(err);
+                        }
+                    );
+                })
+            );
+        }
+        // updates only for the newly fetched messages
+        return messagePromises.length > 0
+            ? messagePromises[0]
+            : new Promise((resolve, reject) => resolve([]));
```
shadowbas, line 341: No need to do `new Promise...`. The function is async. Just `return []`. The return value will be a resolved promise of an empty array.

Somewhat related, you can create a resolve promise with `Promise.resolve([])`. Not relevant here because the function is async.

src/app/rmmapi/rbwebmail.ts
```diff
@@ -265,6 +273,76 @@ export class RunboxWebmailAPI {
...
+        const cachedIds = await this.checkCachedMessageContents([...messageIds]);
+
+        // Filter out items we already have
+        // or are busy fetching
+        const missingMessages = messageIds.filter(
+            (msgId) => !cachedIds.includes(msgId)
+                && !this.downloadingMessages.includes(msgId)
+        );
+ 
+        const messagePromises = [];
```
shadowbas, line 287: messagePromises array functions somewhat like a flag.

src/app/rmmapi/rbwebmail.ts
```diff
@@ -265,6 +273,76 @@ export class RunboxWebmailAPI {
...
+        if (missingMessages.length > 0) {
+            this.downloadingMessages = this.downloadingMessages.concat(missingMessages);
+            messagePromises.push(
+                new Promise((resolve, reject) => {
```
shadowbas, line 291: I think it would make sense to remove the messagePromises and do a return early here. Then at the other return it returns an empty array.

shadowbas (approved):
I like that you placed the message processing in its own method. The rest looks good.

Issue #1095, opened 2021-10-06
#Mail: Drag-and-drop of messages shows the wrong entry


gtandersen:
**Describe the bug**
Drag-and-drop of messages shows the wrong entry being moved, and the entry is incorrectly rendered.

In Firefox and Chrome the entry shown are 2-3 entries further up in the list, and split horizontally as a combination of two entries. 

In Safari the symptom is similar, but the shown entry is enlarged to double the size.

**To Reproduce**
1. Go to https://runbox.com/app
2. Click and drag a message in the message list
3. Inspect the entry being dragged

**Expected behavior**
The correct entry should be shown when dragging, and not split horizontally.

**Screenshots**
Screenshots unavailable due to the symptom only being shown when clicking and dragging.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Firefox, Chrome, Safari
 - Version: Most recent

castaway:
The current live system seems to work fine for me, using Firefox on linux, Chrome-based browsers however, don't support the method we're using (it will only drag with a visible image) - I've managed to sort of make that work..

gtandersen:
Ref https://github.com/runbox/runbox7/commit/5ffcaf29f8c01c6c8c233ef75793d849922d7be9

gtandersen:
It might be a good idea to only use a visible image in all browsers as it might be more reliable.

gtandersen:
Related to #863.

PR #1651, opened 2025-01-21
fix(drag): Shows correct drag-image while moving images to folders

castaway:
This was "fun with browsers and standards" time, chrome needs the canvas-to-copy to be in the DOM (MDN/firefox doesnt), either way whats used as the dragimage is the, view of the element as the user would see it.. so we were seeing whatever  was displaying where the dragcanvas element was being created (which was the top of the main canvas).

Fixed by creating the canvas-copy in the code, and adding it to the DOM off the left edge (off the top doesnt work in all browsers... standards!) 

Fixes: #1095

### DIFF
--- a/src/app/canvastable/canvastable.component.html
+++ b/src/app/canvastable/canvastable.component.html
@@ -15,4 +15,3 @@
             (dragstart)="dragColumnOverlay($event)"
             >              
 </div>
-<canvas id="thedragimage"></canvas>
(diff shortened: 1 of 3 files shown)


shadowbas (approved):
Seems good. This code will be dropped if accessible table gets merged.

runboxdave:
The drag image looks incorrect to me. It seems to have the top half of one message row and the bottom of another.

castaway:
@runboxdave 's report is referring to macos/retina screens (sigh) refixed in #1657

Issue #1559, opened 2024-04-20
#Mail: Message action buttons should only be enabled when a message is selected


gtandersen:
**Describe the bug**
When clicking any of the message action buttons without one or more messages being selected, no action is performed or an error is produced. The buttons should only be enabled when one or more messages are selected.

**To Reproduce**
1. Go to https://runbox.com/app
2. Click on e.g. the "Trash" or "Mark unread" button.
3. Observe that no action is performed or an error is produced.

**Expected behavior**
The message action buttons should be disabled and greyed out, and not be clickable, until at least one message has been selected.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Safari

PR #1652, opened 2025-01-22
fix(messagelist): Only show message action buttons when messages are …

castaway:
…selected

Fixes: #1559


shadowbas (approved):

runboxdave:
Seems to work as described.

PR #1653, opened 2025-01-22
task(angular): Upgrade from 15 to 16

shadowbas:
- [ ] Replace outdated angular2-qrcode package with something else (https://www.npmjs.com/package/angular2-qrcode?activeTab=versions is 6 years old)

Besides that the build works and the app seems to work.


shadowbas:
Should have been a draft PR

PR #1681, opened 2025-05-15
Make folder buttons keyboard navigable

shadowbas:
It is currently not possible to select a folder using the keyboard. Considering this is a primary use-case I decided to fix it as one of the first a11y improvements.

### DIFF
--- a/src/app/folder/folderlist.component.scss
+++ b/src/app/folder/folderlist.component.scss
@@ -1,7 +1,28 @@
-.mailFolder:hover {
+.mailFolder {
+    position: relative;
+}
+
+.mailFolder:hover,
+.mailFolder:focus-within,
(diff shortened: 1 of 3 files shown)


src/app/folder/folderlist.component.scss
```diff
@@ -1,7 +1,28 @@
...
-}  
+}
+
+.mailFolder a {
+    margin-left: 1ch;
+    text-decoration: none;
+    color: #444;
+}
+
+.mailFolder a::after {
```
shadowbas, line 18: A common trick to make the parent of the anchor also clickable.

It is not allowed to have buttons and such within an anchor. That is why one has to opt for this trick.

castaway (approved):
lgtm

Issue #1510, opened 2023-11-12
#Feature-Add: Email Validation


gtandersen:
List of signup related issues in process order:

- [x] Decrease size of initial index sync modal that currently overlays the welcome offer: https://github.com/runbox/runbox7/issues/1509 
- [x] Ensure welcome email is visible in the Inbox upon signing up
- [ ] Clarify that the alternative email address must be validated on https://runbox.com/app/welcome?postSignup=offer (although an email is automatically sent to the submitted address)
- [ ] No indication on https://runbox.com/app/account/details or https://runbox.com/mail/account that the alternative email address has been validated
- [ ] No information or link to validate alternative email address from https://runbox.com/app/account/details
- [ ] The email validation link that is sent takes the user to Runbox 6 instead of 7
- [ ] Show a warning upon login if the alternative email address has not been validated
- [ ] Accept validation without a logged-in session, to prevent errors when users attempt this and receive an error.

castaway:
Patch to fix the welcome email issue submitted (to backend code)

castaway:
See #1531 for item 1

orangesunny:
I can confirm that it is still impossible to finish _Alternative email address_ validation in either Runbox 7 or 6. The link arrives to alternative inbox quickly, but opening the link shows ✔ next to the _Alternative email address_ field only very shortly. It asks for a password in a dialog/pop-up window. No matter what the user does after, the check disappears, validation is still unfinished and requested in _Account_ details as well as **IN ANNOYING RED TEXT** on the home screen of RB7.

Tested in Firefox and de-googled Chromium on macOS 15.2

PR #1690, opened 2025-06-02
fix(details): show alternative email resend on email change

shadowbas:
Issue was likely caused by an error that was thrown between the default true value of the `is_alternative_email_validated` and the request response callback.

- Decided to have as default that the alt email was not validated.
- Added a null check that was likely the reason for the error.

#1510 specifically https://github.com/runbox/runbox7/issues/1510#issuecomment-2697748691


castaway (approved):
lgtm

PR #1701, opened 2025-06-16
fix(login): make username and password required

shadowbas:
Remove the empty redundant ngModel directive.

### DIFF
--- a/src/app/login/login.component.html
+++ b/src/app/login/login.component.html
@@ -24,13 +24,13 @@
 	  <p>Log in below or <a href="/signup?runbox7=1" class="login-link">create a new account</a>.</p>
 	</div>
 	<mat-form-field>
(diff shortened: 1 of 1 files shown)


castaway (changes requested):
It looks like validation is supported in template-driven forms, and as we don't need to capture valuechanges or similar, I think this would be better in KISS style by just adding "required" to the input fields, rather than switching to Reactive forms: https://v16.angular.io/guide/form-validation

shadowbas:
I guess, what I said is not completely correct. The main issue really is that one has to render the errors themselves using mat-error component. The reactive approach has the rendering of errors baked in.

shadowbas:
@castaway , you are correct that template driven forms offer ability to perform required validation. I read that forms with validation should use reactive forms and went with that.

shadowbas:
@castaway, rebased and ready to merge.

PR #1703, opened 2025-06-19
fix(notification): notification permission on click

shadowbas:
Perform notification permission request also on
user click.

Include a button for requesting permissions.


castaway (commented):
Looks fairly sane to me, it seems to repeat the mat-item/button changes in another PR, was that on purpose?

Could you include a screenshot with the button so that @gtandersen can review? (I'm wondering if the button should go on the top bar. instead of the mail sidebar..)

PR #1725, opened 2025-10-05
style(payment): Replace PayPal with Stripe, add new payment method logos, and improve layout.

gtandersen:
Payment methods


src/assets/payment/Credit_Cards_PayPal.png
castaway, line 1: Should this one include Apple / Google Pay ? (also available via Stripe)

Issue #1399, opened 2023-03-30
Calendar "add event" button missing


kotecky:
**Describe the bug**
The (+) button is available in in the calendar on hover over a date: 
![image](https://user-images.githubusercontent.com/2297444/228777865-cbb70fef-307e-4736-9dee-4f8caced56dd.png)

It should (but is not) the case in the same browser window, just made smaller and where the app adapts to a smaller size. 
![image](https://user-images.githubusercontent.com/2297444/228777551-d3d5393c-25fa-474b-9f7c-f4210e0b75a8.png)

A page refresh does not change the situation. 

Similarly on a mobile (Android8, FF) the above way of adding events is missing. On purpose? 

**To Reproduce**
1. Go to calendar
2. Hover on days
4. See (+) button missing for windows 

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser: FF 111

**Smartphone (please complete the following information):**

**Additional context**

PR #1731, opened 2025-10-13
fix(calendar): Show add button on day in month view on narrower screens

shadowbas:
Issue #1399


castaway (changes requested):
Hi @shadowbas in mobile mode this makes the calendar view very squished (since I think you've just removed the special mobile view).. and the + button doesnt display by default, only if you "mouse over" the corner its supposed to be in.. which isnt possible on mobile

2025-12-03 15 45 12 mallet3 scsys co uk 81e683ad8ef5

PR #1752, opened 2025-12-11
fix(notifications): Check Notifcation API exists when showing button

castaway:
Fix: "blank screen" on iPad/iPhones. Notification API does not exist on Safari iOS


antoniskalou (approved):
LGTM

Issue #939, opened 2021-03-23
Addresses in the body of emails are made in to mailto: links


davidbowdley:
Addresses in the body of an plain text emails are converted to mailto: links. 

If the addresses are to be links they should open a compose window and not activate an external email client.

PR #1754, opened 2026-01-03
fix(mailviewer): intercept `mailto:` links and open a compose window

gacevedo:
Intercept `mailto:` links and open a compose window. Fixes issue #939.

### DIFF
--- a/src/app/mailviewer/singlemailviewer.component.spec.ts
+++ b/src/app/mailviewer/singlemailviewer.component.spec.ts
@@ -17,7 +17,7 @@
 // along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
 // ---------- END RUNBOX LICENSE ----------
 
-import { ComponentFixture, TestBed, tick, fakeAsync, waitForAsync } from '@angular/core/testing';
+import { ComponentFixture, TestBed, tick, fakeAsync, waitForAsync, flush } from '@angular/core/testing';
 
 import { SingleMailViewerComponent } from './singlemailviewer.component';
 import { ResizerModule } from '../directives/resizer.module';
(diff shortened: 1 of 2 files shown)


gacevedo:
> @castaway I might need help defeating that `Commitlint` check. The documentation might be outdated or I'm not understanding something: https://github.com/runbox/runbox7/blob/master/CONTRIBUTING.md#revert.
> 
> > If the commit reverts a previous commit, it should begin with revert:, followed by the header of the reverted commit. In the body it should say: This reverts commit ., where the hash is the SHA of the commit being reverted.
> 
> Please advise. Thank you!

Never mind. I just deleted the commit(s) instead.

juanpeha:
@gacevedo, this looks like a nice starting point and the core interception logic works well.

  I have some questions about the implementation:

  Performance & Lifecycle:
  - Is there a reason for calling initMailtoInterceptor() in ngDoCheck()? This would remove/re-add event listeners on every change detection cycle - could this cause performance issues?
  - Should there be cleanup in ngOnDestroy() to prevent memory leaks when the component is destroyed?
  - Is the setTimeout wrapper in the constructor necessary? Does ngAfterViewInit not fire after the view is ready?

  Functional Questions:
  - The regex only extracts the address part before ? - should it also parse query parameters like cc, bcc, subject, body?
  - How should mailto:?to=alice@example.com be handled (where to is in the query params)?
  - What happens with empty query values like mailto:first@example.com?to=? Should these be trimmed?
  - Since messageContents is behind *ngIf="mailObj" and gets recreated when switching messages, how does the interceptor reinitialize for the new element?

  Code Quality:
  - Lines 47-59 have duplicate afterAll blocks - was this intentional?
  - Could _mailtoInterceptorListener be more specifically typed instead of any?
  - Should there be test coverage for the additional mailto parameters?

PR #1760, opened 2026-01-23
fix(compose): add double line break at top of replies

RunboxJoe:
When replying to messages, add a double line break at the top of the draft.

resolves 911


antoniskalou:
So is this good to go @castaway @RunboxJoe?

PR #1768, opened 2026-02-05
style(searchservice): change of wording to be more clear to the user what they are downloading

RunboxJoe:
Previously the pop-up gave incorrect information to the user. It asked if the user wanted to
synchronize entire index of x number of messages where x was actually the additional number
of messages yet to be synchronized.
Message now correctly refers to x as the additional number of messages
Commented variable declaration of variable that may have caused this confusion due to its name


castaway (approved):
lgtm

PR #1778, opened 2026-03-23
fix(account-details): validate personal name fields

### LINKED ISSUE #1739: #Signup First and last name fields accept non-word characters
**Describe the bug**
When signing up for an account, the first and last name fields accept non-word characters and therefore strings such as links. These can then be rendered as clickable links in the alternative email address verification email that is sent.

**To Reproduce**
1. Go to https:/runbox.com/signup
2. Enter e.g. domain.tld in the first or last name field, and complete the rest of the form
3. In the alternative email address verification email that is sent, the above string is shown as a link.

**Expected behavior**
Only word characters that would normally appear in names should be accepted.

 - OS: macOS
 - Browser: Firefox

snorrea:
Reject URL-like strings in personal details before submitting account updates. Normalize whitespace and show validation errors so invalid names cannot be saved.

Closes #1739

### DIFF
--- a/src/app/account-details/personal-details.component.html
+++ b/src/app/account-details/personal-details.component.html
@@ -11,10 +11,16 @@
                 <mat-form-field>
                     <mat-label>First name</mat-label>
                     <input type="text" matInput formControlName="first_name">
+                    <mat-error *ngIf="detailsForm.get('first_name').hasError('invalidName')">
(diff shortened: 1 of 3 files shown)


castaway:
Hi,

Thanks for the contrubution! Sadly this won't fix #1739 as the signup form isn't in this repository, however it should improve the account page

PR #1779, opened 2026-03-23
fix(calendar): use toJSDate() for proper timezone conversion

piratefinn:
Fixes timezone conversion bug in calendar events where events were displayed at incorrect times.

The previous implementation used `new Date(string)` which parses the date string as local browser time, leading to incorrect timezone conversions. For example, an event at 09:00 Berlin (CEST=UTC+2) was incorrectly converted to local time based on browser timezone instead of properly representing the absolute UTC time.

ICAL.Time provides `toJSDate()` which correctly handles the timezone conversion from ICAL time to JavaScript Date while preserving the correct UTC time.

## Changes

- `runbox-calendar-event.ts`: Changed `new Date(string)` to `toJSDate()` for both `start` and `end` getters
- Updated tests to reflect correct timezone behavior


src/app/calendar-app/runbox-calendar-event.spec.ts
```diff
@@ -375,4 +374,237 @@ END:VCALENDAR`
...
+
+        // Same floating time for Berlin user - interpreted as Berlin time (UTC+1) = 15:00 UTC
+        const floatingEventBerlin = new RunboxCalendarEvent(
+            'testcal/floating-berlin',
+            new ICAL.Event(new ICAL.Component(['vevent', [
+                ['dtstart', {}, 'date', '2026-03-24T16:00:00'],
+                ['dtend', {}, 'date', '2026-03-24T17:00:00'],
+                ['summary', {}, 'text', 'Floating event at 4pm (Berlin user)'],
+            ]])),
+            ICAL.Time.fromDateTimeString('2026-03-24T16:00:00'),
```
castaway: Can we use icalevent.startDate like https://github.com/runbox/runbox7/blob/adfbf32b414d300c47096fbf5c8c9b06864a6bd5/src/app/calendar-app/calendar.service.ts#L380 - instead of creating fresh ICAL.Times here? that way we're testing what actually happens in the code (aka how RunboxCalendarEvent is used

castaway (approved):
Looks good to me! (Aren't timezones fun..)

piratefinn:
# Testing Checklist

## Calendar Events

- [ ] Events with a set timezone (e.g. "14:00 Oslo") show at the correct time in your local timezone
- [ ] A 4pm London event appears as 5pm if your timezone is Berlin, **not** 3pm (the original bug)
- [ ] Events with UTC time (ending in Z) show at the correct local time
- [ ] Events without a timezone are treated as being in your calendar's timezone

## All-Day and Recurring Events

- [ ] Multi-day all-day events span the correct days (not shifted by one day)
- [ ] Recurring events show every occurrence at the right time, not just the first one

## Importing Calendar Files (.ics)

- [ ] Importing a .ics file shows events at the correct times in the preview
- [ ] Importing a .ics file with multiple events imports all of them
- [ ] After import, events appear at the correct times in month, week, and day views

## Editing Events

- [ ] Changing an event's time and saving works correctly
- [ ] Moving one instance of a repeating event changes only that instance

## Timezone Changes

- [ ] Change your computer/browser timezone to a different one (e.g. New York) and check that events shift to match
- [ ] Events around daylight saving time changes (last Sunday of March/October) display correctly

## Other Features Still Work

- [ ] Email list loads and displays normally
- [ ] Composing and sending an email works
- [ ] No error messages appear in the browser console while using the calendar

runboxdave:
Doing some testing here:

1. Added an event to start at 12pm but it appears as an 11am event (1 hour earlier)
2. An all day event appears on the day before it was set to appear but later as the calendar refreshes moves to the correct day.
3. The same as in 2. above happens for multi-day events.

PR #1780, opened 2026-03-25
fix(compose): close CC/BCC fields when clicking close button (#1776)

### LINKED ISSUE #1776: CC/BCC close button does not close the respective fields
**Describe the bug**
When the CC or BCC fields get open by clicking the CC/BCC buttons, the respective field opens together with a cross-like button meant to close the field if not needed. Clicking on the cross-like button does not close the CC/BCC fields, which would be the expected behaviour. 

Image

 - OS: Win11
 - Browser: FF
 - Version: 148.0.2

Keegil:
Fixes #1776

## Problem
When clicking the close (X) button on CC/BCC fields, the form control value was cleared but the fields remained visible because `hasCC`/`hasBCC` were not toggled back to `false`.

## Fix
Updated both close button click handlers to:
1. Clear the form control value (`setValue(null)`)
2. Set `hasCC`/`hasBCC` back to `false` to hide the field
3. Clear the recipients array (`model.cc.splice(0)` / `model.bcc.splice(0)`)

## Testing
- Verified build compiles cleanly (`ng build runbox7`)
- Manual verification: clicking X on CC/BCC fields now properly hides them

Closes #1776

### DIFF
--- a/src/app/compose/compose.component.html
+++ b/src/app/compose/compose.component.html
@@ -89,7 +89,7 @@
                 [recipients]="model.cc"
                 (updateRecipient)="onUpdateRecipient('cc', $event)"
             ></mailrecipient-input>
-            <button mat-icon-button (click)="formGroup.controls.cc.setValue(null)"><mat-icon svgIcon="close"></mat-icon></button>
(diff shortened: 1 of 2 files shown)


gtandersen:
@Keegil Many thanks for these fixes and for clearly and concisely describing them. We will review and comment in kind!

PR #1781, opened 2026-03-25
fix(compose): remove signature when switching to identity without one (#1572)

### LINKED ISSUE #1572: #Compose Switching to an Identity with no signature should remove previous signature
**Describe the bug**
When starting to compose an email, if the default identity has a signature, we insert it. If we then switch to one with no signature, the original signature should be removed

**To Reproduce**
1. Make sure the default identity has a signature
2. Go to Compose
3. Choose another identity with no signature
4. Original signature should be removed

Keegil:
Fixes #1572

## Problem
When composing an email with a default identity that has a signature, switching to an identity with no signature left the original signature in the message body. The code handled three cases:
1. Both old and new identity have signatures → replace
2. No old signature, new identity has one → prepend
3. ❌ Old signature exists, new identity has none → **nothing happened**

## Fix
Added the missing third case: when the current identity has a signature and the new one does not, the old signature is removed from both the plain text body and HTML content.

The regex properly escapes special characters in the signature text to avoid regex injection issues.

## Testing
- Verified build compiles cleanly (`ng build runbox7`)

Closes #1572

### DIFF
--- a/src/app/compose/compose.component.spec.ts
+++ b/src/app/compose/compose.component.spec.ts
@@ -1,0 +1,50 @@
+// --------- BEGIN RUNBOX LICENSE ---------
+// Copyright (C) 2016-2022 Runbox Solutions AS (runbox.com).
+//
+// This file is part of Runbox 7.
+//
+// Runbox 7 is free software: You can redistribute it and/or modify it
+// under the terms of the GNU General Public License as published by the
+// Free Software Foundation, either version 3 of the License, or (at your
+// option) any later version.
+//
+// Runbox 7 is distributed in the hope that it will be useful, but
+// WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+// General Public License for more details.
+//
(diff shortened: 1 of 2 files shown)


src/app/compose/draftdesk.service.spec.ts
castaway, line 1: These tests just test "how regex replace works", they won't fail if someone changes the code. Better tests would be: in the compose component, extract the "changes signature" section to a javascript function, then in the tests - test that function.

castaway (changes requested):
See notes - also needs rebasing on the recently updated master branch

Keegil:
Thanks, I extracted the signature-switching logic into a helper in `compose.component.ts` and rewrote the regression coverage to test that helper directly, then pushed the update. I also verified the branch is current with `master` and re-ran local TypeScript checks (`tsc -p src/tsconfig.app.json --noEmit` and `tsc -p src/tsconfig.spec.json --noEmit`).

Keegil:
Rebased this branch on the current master and force-pushed the update. I also re-ran local TypeScript compile checks (src/tsconfig.app.json and src/tsconfig.spec.json).

PR #1791, opened 2026-03-26
Fix #406: keep spam messages in plain view by default

### LINKED ISSUE #406: When turning on HTML viewing by default, don't apply this setting to the Spam folder
I've turned on the HTML viewing for my messages so that by default I see formatted text and images. 

However, I don't think this setting should be applied to messages in the spam folder. I went there to review my spam and was surprised to see images load (as I don't normally have that enabled for spam) and then realised it was because I had turned on HTML viewing.

For the obvious reasons it isn't desirable to download remote content in spam messages by default.

Keegil:
## Summary
- ignore the global always-show-HTML preference for messages in the Spam folder
- set the message folder before applying sender-based HTML preferences
- add focused regression coverage for spam and non-spam folders

## Testing
- npm run lint *(project has pre-existing warnings only)*
- npx ng test runbox7 --watch=false --browsers=ChromeHeadless --include=src/app/mailviewer/singlemailviewer.component.spec.ts *(fails on this host because ChromeHeadless is not installed)*
- npx ng build runbox7

Fixes #406


castaway (approved):
Thanks for the patches! It may take us a bit to go through them and review

castaway:
hi @Keegil, no need to add the changelog, as the build script does that these days (I have a PR waiting to remove that instruction) - so please remove it again

Keegil:
Good catch — removed the changelog entry. The build script will handle it. Thanks!

PR #1802, opened 2026-03-27
Fix #1753: show embedded images without external image toggle

### LINKED ISSUE #1753: Show External Images option is required even for embedded images
**Describe the bug**
Show External Images option is required even for embedded images.

**Expected behavior**
Turning on the HTML option should be enough in itself to display embedded images.

Keegil:
**Bug fix** — Embedded (CID-backed) images are hidden unless "Show External Images" is enabled, even though they are not remote resources.

## Problem
Turning on HTML view was not sufficient to display inline images attached via Content-ID (CID). Users also had to enable "Show External Images" — a toggle intended for remotely-hosted images — before embedded images would appear. This was incorrect behaviour: CID images are part of the message and carry no privacy risk. Issue #1753.

## Fix
- Show CID-backed inline images as soon as HTML view is enabled when the message contains no truly external image sources
- Continue blocking remote images by default when external image URLs are present, preserving the existing privacy protection

## Testing
- Regression tests added for inline-only and mixed inline/external image messages in `singlemailviewer.component.spec.ts`
- `npx tsc -p src/tsconfig.spec.json --noEmit` passes
- `npx tsc -p src/tsconfig.app.json --noEmit` passes

Fixes #1753

### DIFF
--- a/src/app/mailviewer/singlemailviewer.component.spec.ts
+++ b/src/app/mailviewer/singlemailviewer.component.spec.ts
@@ -231,6 +231,62 @@ describe('SingleMailViewerComponent', () => {
       expect(component.mailObj.attachments[1].downloadURL.indexOf('blob:')).toBe(0);
     }));
 
+  it('keeps inline images visible without enabling external images', () => {
+    component.messageId = 77;
+
+    const processed = component['processMessageContents'](Object.assign(new MessageContents(), {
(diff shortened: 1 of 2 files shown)


castaway:
This issue cannot be fixed with a frontend change - it is a backend bug

PR #1804, opened 2026-03-27
fix(mail): close preview pane on second subject click

### LINKED ISSUE #1450: Close preview pane by clicking on subject a second time
When the preview pane is open, a second click on the same message could close the pane.

Keegil:
## Summary
- close the preview pane when the currently opened subject row is clicked a second time
- add a regression spec covering the MessageDisplay row selection behavior
- update the generated changelog entry for issue #1450

## Testing
- npm run lint
- node src/build/gen-env.js
- npm run test -- --watch=false --progress=false --browsers=FirefoxHeadless --include src/app/common/messagedisplay.spec.ts
- git log -1 --stat --oneline

## Results
- `npm run lint` passed
- the targeted Karma test gets past TypeScript compilation after generating `src/environments/env.ts`, then stops because this machine does not have a Firefox binary for `FirefoxHeadless`
- `git log -1 --stat --oneline` confirms the published fix commit

Closes #1450


castaway (approved):
Did not need changes.ts - rest looks good

castaway:
Hi @Keegil I have reverted this as the test (and change!) did not actually affect the mail pane, "openedRowIndex" is used for highlighting the correct row in the message list (and ensuring we are scrolled to it if necessary) the mail pane open/close does not use it

PR #1806, opened 2026-03-27
fix(calendar): restore week navigation for saved week view

### LINKED ISSUE #1541: Previous and next button in calendar will jump next/previous month instead of week.
**Describe the bug**
While my calendar display setting is 'week',  when I click on 'next' or 'previous', it will jump to next or previous month.
ie: when displaying current week (26/02 to 03/03), clicking on next will jump to week 1/4 to 7/4.

**To Reproduce**
1. Go to calendar
2. Click on 'next'
3. See error

**Expected behavior**
While my display setting is 'week', I would expect 'next' and 'previous' button to jump to next or previous week

 - OS: [win10]
 - Browser [chrome]
 - Version [121]

Keegil:
**Bug fix** — Previous/Next navigation in the calendar advances by month instead of week when the saved view preference is Week.

## Problem
When a user's calendar view preference was saved as Week, the toolbar's Previous/Next buttons navigated by month instead of week after the app reloaded. The Angular Calendar directive `mwlView` stayed at its default `Month` value because `setView()` only updated it when the stored preference changed — it did not apply the initial restored value. Issue #1541.

## Root cause
`setView()` compared the new value against the current `view` property. When the app restored `lastUsedView = Week` from preferences, `view` became `Week`, but since `setView()` treated this as "no change", `mwlView` was never updated from its default `Month`. The toolbar directives read `mwlView` for navigation step size, so all navigation used month steps.

## Fix
- Synchronise `mwlView` whenever the calendar view is restored from preferences or changed by the user
- Ensure the initial preference restoration sets both `view` and `mwlView` consistently

## Testing
- Regression spec added: restores week view from preferences and verifies the Next button advances by one week
- `npm run lint` passes (existing repo-wide warnings only)

Closes #1541


src/app/changelog/changes.ts
castaway, line 1: Remove the changes.ts please

src/app/calendar-app/calendar-app.component.ts
```diff
@@ -330,28 +330,29 @@ export class CalendarAppComponent implements OnDestroy {
 
     setView(view: RunboxCalendarView): void {
         this.view = view;
+
```
castaway: I disagree with the the root cause assessment here, the issue lies with the preferences restoration, which sets the settings to the preferences, then calls setView(). I suggest keeping setView as it was, and changing the preferences restore to be smarter. It could even save/restore currentView/mwlView instead of lastUsedView, then it would Just Work

castaway (changes requested):

Keegil:
Updated this to keep setView() unchanged and instead restore the saved calendar view directly from preferences, plus removed the changes.ts update. I also re-ran a TypeScript compile check locally.

PR #1808, opened 2026-03-27
fix(mailviewer): add origin time toggle for opened messages

### LINKED ISSUE #1455: Show time zone option for sent time in opened messages
It could be useful with an option that gives the user a way to choose how the sent timestamp should be displayed.

This can be implemented as a button that says either "Local time" or "Origin time" depending on the mode, and that switches between the two modes where local time does not show an offset but the origin time does (e.g. +01:00).

See GL-RMM 962 for more details.

Keegil:
## Summary
- add a compact time-mode toggle beside the opened-message timestamp
- keep the existing local-time display as the default view without an offset
- show the original message timestamp with its source offset when the user switches to origin time
- add a regression spec covering the toggle in `SingleMailViewerComponent`
- update the generated changelog

## Implementation notes
The message viewer already normalizes the displayed timestamp for local display. This change preserves that default behavior and adds a separate origin-time rendering path that formats the original message header timestamp directly, including its offset.

## Testing
- `npm run lint`
- `node src/build/gen-env.js`
- `npm run test -- --watch=false --progress=false --browsers=FirefoxHeadless --include src/app/mailviewer/singlemailviewer.component.spec.ts`
- `node src/build/build-changelog.js`
- `git log -1 --stat --oneline`

## Results
- `npm run lint` passed with the repo's existing warnings only
- the targeted mail viewer spec starts Karma successfully and then stops because this machine does not have a Firefox binary for `FirefoxHeadless`
- `git log -1 --stat --oneline` confirms the published fix commit

Closes #1455


Keegil:
Closing — the origin-time toggle feature is valid, but the diff inadvertently removes timezone from the default time display, which is an unintended regression. Will revisit if I can separate the two changes cleanly.

PR #1809, opened 2026-03-27
fix(compose): preserve reply and forward subjects

### LINKED ISSUE #1093: Stop prepending reply and forward subjects with Re or Fwd
It was the culture in the past to prepend reply and forward subjects with Re or Fwd. It results in `Re:RE:Re:...` in the better case, `Re:AW:Re:...` otherwise. Many (most?) desktop clients or online services do not do it any more.  Probably it is linked with a international environment that we work-live in. 

Consider to stop changing the subject text in a reply or forward.

Keegil:
## Summary
- preserve the original message subject when creating reply and forward drafts
- update compose unit coverage for plain, nested, and already-prefixed subjects
- update mailviewer Cypress expectations so compose no longer requires injected Re:/Fwd: prefixes

## Verification
- `npx ng test --watch=false --browsers=FirefoxHeadless --include src/app/compose/draftdesk.service.spec.ts` *(blocked locally: FirefoxHeadless binary is not installed in this environment)*
- `npx eslint src/app/compose/draftdesk.service.ts src/app/compose/draftdesk.service.spec.ts`
- `npm run build`

## Issue
Closes #1093

### DIFF
--- a/e2e/cypress/integration/mailviewer.ts
+++ b/e2e/cypress/integration/mailviewer.ts
@@ -55,7 +55,8 @@ describe('Interacting with mailviewer', () => {
             expect(loc.pathname).to.eq('/compose');
         });
         cy.wait(500);
-        cy.get('mat-card-actions div').should('contain', "Re: No 'To', just 'CC'");
+        cy.get('mat-card-actions div').should('contain', "No 'To', just 'CC'");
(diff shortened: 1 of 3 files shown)


Keegil:
Closing — this incorrectly strips Re:/Fwd: prefixes from reply and forward subjects. That's a regression, not a fix. Apologies for the noise.

PR #1823, opened 2026-04-15
Re-enable user actions (angular v16/eslint update broke some xapian things subtly)

castaway:
Does this fix the user actions issues? Maybe
Does it pass tests? Not on my machine..

After a stern talking to, passes tests and user actions work !

### DIFF
--- a/e2e/cypress/integration/account-access-subaccount.ts
+++ b/e2e/cypress/integration/account-access-subaccount.ts
@@ -1,0 +1,40 @@
+describe('Account access control', () => {
+    function becomeSubaccount() {
+        cy.intercept('/rest/v1/me', (req) => {
+            req.continue((res) => {
+                const payload = JSON.parse(res.body);
+                payload.result.owner = {
+                    uid: 666,
+                    username: 'mastermind@runbox.com',
+                };
+              res.send(JSON.stringify(payload));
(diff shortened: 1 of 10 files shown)


piratefinn (approved):
Boop.

piratefinn:
The xapian revert makes sense to me. The Sentry PR (50b8e5db) switched `declare var` to `declare let` and changed some `.apply()` calls to use spread syntax, and this just puts it all back to the state that was working. The `Observable<boolean>` tightening on `deleteLocalIndex` is a nice bonus over what master had.

ESLint approach is a bit inconsistent (blanket `/* eslint-disable */` in the two loader files vs targeted `// eslint-disable-next-line no-var` in the other three) but totally fine given the loader files already had a stack of rule disables.

E2E changes look good. Adding `cy.intercept` + `cy.wait` before assertions is definitely the right move over relying on 10s timeouts. Splitting the subaccount tests out into their own file keeps things tidy. The mockserver receipt endpoint fix is a reasonable defensive change too.

Minor thing: double space in `declare var  IDBFS;` in both loader files, and the PR title says "revert(index)" but it touches the loader/patching code as well. Neither worth changing.

piratefinn (approved):

PR #1839, opened 2026-05-13
Use app favicon

### LINKED ISSUE #922: web app: favicon
A tiny non-issue: the favicon of the web app (https://runbox.com/_img/index/logobox.png) is different from the current Runbox logo. On purpose?

Wheezily:
Fixes #922.

The app shell was still loading the favicon from the old site logo at `/_img/index/logobox.png`. This points the page at the app favicon instead and adds `src/favicon.ico` to the Angular build assets so the file is emitted.

Validation:
- `node -e "const fs=require('fs'); JSON.parse(fs.readFileSync('angular.json','utf8')); if(!fs.existsSync('src/favicon.ico')) process.exit(1); console.log('angular.json ok; favicon exists');"`
- `git diff --check`

I did not run a full Angular build locally because dependencies are not installed in this checkout.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -17,6 +17,7 @@
             "tsConfig": "src/tsconfig.app.json",
             "polyfills": "src/polyfills.ts",
             "assets": [
+              "src/favicon.ico",
               "src/assets",
               {
                 "glob": "xapianasm*",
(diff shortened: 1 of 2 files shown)


Wheezily:
Added a fuller local validation pass:

- Installed dependencies in a Node 18 container with the Cypress binary download disabled.
- Fetched full git history so the changelog step can resolve its base commit.
- Ran `npm run build`; it passed with the existing Angular/CommonJS/Sass warnings.
- Re-ran `git diff --check`; it passes.

PR #1827, opened 2026-05-05
feat(signup): Angular version of signup page using existing backend.

gtandersen:
This is an AI assisted attempt at upgrading the Runbox 6 signup screen to Runbox 7 in Angular without modifying the backend except for Apache rewriting. Fixes #1354.

**Summary**
Runbox 7 signup has been implemented as a new Angular route and is now functionally working against the existing signup backend. The page is live in the test environment, visually modernized, responsive, and compatible with the current Apache/nginx routing setup.

Runbox 7 signup page

<br />

**What Has Been Delivered**
- New Angular signup screen at /app/signup
- Compatibility with existing legacy signup processing
- Apache rewrite support for /signup -> /app/signup
- Deep-link compatibility for Angular routing under /app
- Responsive layout for desktop, tablet, and mobile
- Updated UI aligned with Runbox 7 visual style
- Inline form validation feedback
- hCaptcha integration
- Dynamic loading of:
  - signup form action
  - hCaptcha site key
  - Runbox domain list

**How It Works**
The Angular page does not yet use a dedicated signup metadata API. Instead, it fetches the existing legacy signup page and parses it to obtain:
- form action
- available Runbox domains
- hCaptcha site key

Form submission is then handed off to the existing backend signup flow using the same field names and POST behavior expected by the legacy system.

**Backend Impact**
No new signup backend business logic was required.

However, this implementation does depend on the existing backend providing:
- reachable legacy signup HTML
- stable legacy form markup
- working hCaptcha/backend services

So this is not purely frontend-only in operational terms, even though no new signup processing code was added.

**Testing Status**
Implemented:
- unit test coverage for signup component behavior
- Cypress smoke coverage for deployed page behavior/content
- reusable Firefox-based signup unit test command

Unit test coverage now includes:
- legacy metadata loading
- query-param initialization
- metadata fetch fallback behavior
- field-level validation behavior
- invalid custom domain handling
- CAPTCHA-gated submission
- successful native submit path

What has been validated:
- targeted Firefox unit test passes in headless server setup
- live manual testing confirms page rendering and successful signup flow

What is still environment-dependent:
- Cypress against gw01test can only run from a machine that can actually reach that hostname
- full project CI-equivalent verification still depends on running:
  - npm run lint
  - npm run policy
  - broader CI/test phases as needed

**Current Risks / Technical Debt**
Main technical debt:
- frontend currently parses legacy HTML for signup metadata instead of consuming a dedicated JSON endpoint

Risks:
- coupling to legacy markup structure
[...]

PR #1844, opened 2026-05-27
feat(signup): Angular version of signup page using existing backend.

gtandersen:
Cherry-picked version of https://github.com/runbox/runbox7/pull/1827/ without merges.

Requires GL !871 for URL rewriting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "build": "node src/build/pre-build.js && ng build --configuration production --base-href=/app/ runbox7; RES=$?; node src/build/post-build.js; exit $RES",
     "policy": "node policy-tests/run-all.js",
     "test": "ng test",
+    "test:signup:firefox": "ng test --watch=false --browsers Firefox --include src/app/signup/signup.component.spec.ts",
     "lint": "ng lint",
     "e2e": "ng e2e",
     "start-e2e-server": "start-test mockserver 15000 start-use-mockserver",
--- a/src/app/signup/signup.component.html
+++ b/src/app/signup/signup.component.html
@@ -45,8 +45,8 @@ along with Runbox 7. If not, see <https://www.gnu.org/licenses/>.
           <p>Runbox combines modern webmail, custom domains, standard email access, strong spam filtering, and an emphasis on responsible operations.</p>
         </div>
         <div class="note">
-          <h2>Privacy by business model</h2>
-          <p>Runbox is funded by subscriptions, not advertising, so your email content is not mined to sell attention.</p>
+          <h2>Privacy by Design</h2>
+          <p>You are the customer, not the product. Runbox is paid for by subscribers, not advertisers -- and your data is GDPR protected.</p>
         </div>
         <div class="note">
           <h2>Hosted in Norway</h2>
--- a/src/app/signup/signup.component.html
+++ b/src/app/signup/signup.component.html
@@ -1,0 +1,318 @@
+<div class="signup-shell">
+  <header class="signup-header">
+    <a class="brand" routerLink="/login">
+      <img src="assets/runbox7_negative_white.png" alt="Runbox 7" />
+    </a>
+    <p class="header-tagline">Free 30-day trial of Runbox</p>
+  </header>
+
+  <main class="signup-main">
+    <section class="hero-panel">
+      <div class="hero-copy">
+        <p class="eyebrow">Sustainable and Private Email Hosting</p>
+        <h1>Create a Runbox Account</h1>
+        <p class="hero-text">
+          Sustainable and secure email, your own domain options, and a 30-day free trial with no credit card required.
+        </p>
+        <p class="hero-subtext">
+          Your trial starts immediately. Before it ends, you can choose whether to upgrade and keep using the account.
+          Runbox is a paid email service, not an ad-funded one, and customer email content is private.
+        </p>
+      </div>
+
+      <div class="hero-notes">
+        <div class="note">
+          <h2>Sustainable and secure</h2>
+          <p>Runbox combines modern webmail, custom domains, standard email access, strong spam filtering, and an emphasis on responsible operations.</p>
+        </div>
+        <div class="note">
+          <h2>Privacy by business model</h2>
+          <p>Runbox is funded by subscriptions, not advertising, so your email content is not mined to sell attention.</p>
+        </div>
+        <div class="note">
+          <h2>Hosted in Norway</h2>
+          <p>Your service is operated from Norway by an independent email provider with a long-standing focus on privacy and user control.</p>
+        </div>
+        <div class="note">
+          <h2>How the trial works</h2>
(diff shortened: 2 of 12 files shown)


src/app/aliases/aliases.lister.spec.ts
castaway, line 1: Is this part of signup?

src/app/common/preferences.service.spec.ts
castaway, line 1: Not part of signup

src/app/compose/compose.component.scss
castaway, line 1: Not part of signup

src/app/help/help.component.spec.ts
castaway, line 1: Not part of signup ?

src/app/mailviewer/singlemailviewer.component.spec.ts
castaway, line 1: Not part of signup?

src/app/signup/signup.component.html
```diff
@@ -0,0 +1,362 @@
...
+            </span>
+          </label>
+          <small class="field-error" *ngIf="showFieldError(tosAcceptedModel, signupForm)">You must accept the terms to create an account.</small>
+        </section>
+
+        <p class="form-error" *ngIf="hCaptchaError || submitError" aria-live="polite">
+          {{ submitError || hCaptchaError }}
+        </p>
+
+        <input type="hidden" name="av" value="y" />
```
castaway: Do we still need/use these "av" and "as" values?

src/app/signup/signup.component.html
```diff
@@ -0,0 +1,362 @@
...
+        <input type="hidden" name="as" value="y" />
+        <input type="hidden" name="signup" value="1" />
+        <input type="hidden" name="accountType" [value]="accountType" />
+        <input type="hidden" name="domain" [value]="userDomain" />
+        <input type="hidden" name="account_number" [value]="accountNumber" />
+        <input type="hidden" name="timezone" [value]="timezone" />
```
castaway, line 339: This is not a timezone value.. we should either add a timezone selector, or remove this

src/app/signup/signup.component.html
```diff
@@ -0,0 +1,362 @@
...
+        <input type="hidden" name="runbox7" [value]="runbox7" />
```
castaway: I don't think we use this "runbox7" value at all

src/app/signup/signup.component.ts
```diff
@@ -0,0 +1,377 @@
...
+        const captchaLoaded = await this.loadHCaptchaScript();
+        if (!captchaLoaded) {
+            return;
+        }
+
+        this.hCaptchaReady = true;
+        this.renderHCaptcha();
+    }
+
+    private loadLegacySignupMetadata(): Promise<void> {
```
castaway, line 253: This should not go into production. We know ourselves what the form action is, we do already have a way to get the runbox domains (rbwebmail getRunboxDomains). For the hcaptcha key we can add an API call.

src/app/start/startdesk.component.scss
castaway, line 1: Not part of signup

src/app/websocketsearch/websocketsearch.service.spec.ts
castaway, line 1: Not part of signup

src/app/welcome/welcomedesk.component.spec.ts
castaway, line 1: Not part of signup

src/build/gen-env.js
castaway, line 1: What's this doing? Will supply this via an API.

package.json
```diff
@@ -18,6 +18,7 @@
     "build": "node src/build/pre-build.js && ng build --configuration production --base-href=/app/ runbox7; RES=$?; node src/build/post-build.js; exit $RES",
     "policy": "node policy-tests/run-all.js",
     "test": "ng test",
+    "test:signup:firefox": "ng test --watch=false --browsers Firefox --include src/app/signup/signup.component.spec.ts",
```
castaway, line 21: I would suggest running this manually if wanting to test individual components, I don't think we need a list of them as commands

castaway (changes requested):
Mainly:
* Get rid of the legacy load thing (and use APIs as appropriate)
* Move changes that tweak other non-signup things out of the signup PR
* Review if all the signup specific CSS is needed

src/app/signup/signup.component.html
gtandersen: Isn't this used in Runbox::Webmail::Signup as a timezone string from Intl.DateTimeFormat().resolvedOptions().timeZone? Can be removed if preferable.

src/app/signup/signup.component.html
gtandersen: Removed since not used by the backend signup processing.

src/app/signup/signup.component.html
gtandersen: Removed; they are legacy leftovers and not used by the current signup backend.

src/build/gen-env.js
gtandersen: OK, should I make any changes here?

package.json
gtandersen: OK, should I remove it?

src/app/aliases/aliases.lister.spec.ts
gtandersen: Seemed necessary to make tests pass, but all non-signup commits are now moved to this PR: https://github.com/runbox/runbox7/pull/1850
A clean signup-only PR can be found here to replace the current one: https://github.com/runbox/runbox7/pull/1849

src/app/signup/signup.component.html
castaway: The existing one seems to set it to empty string: `<INPUT type="hidden" name="timezone" value="">` (signup.tpl) which is mildly better, Signup.pm defaults to empty string anyway if not supplied, so might as well drop it (or go the effort to generate one out of the geo::ip data)

package.json
castaway: yes

src/build/gen-env.js
castaway: Yes we dont need the gen-env.js change. Add something to fetch the hcaptcha key similar to the stripe pubkey fetching in payments.service.ts / rbwebmail.ts

src/app/signup/signup.component.ts
gtandersen: Changing it so that it:
- uses rmmapi.getRunboxDomains() for the Runbox domain list
- uses a fixed submit target of signup
- uses a new rbwebmail API method for the hCaptcha site key, following the same frontend pattern as getStripePubkey()

src/build/gen-env.js
gtandersen: OK, changing accordingly.

package.json
gtandersen: OK, removing.

PR #1849, opened 2026-06-02
feat(signup): Angular version of signup page using existing backend.

gtandersen:
Clean version of https://github.com/runbox/runbox7/pull/1844 without unrelated support commits.

### DIFF
--- a/src/app/signup/signup.component.ts
+++ b/src/app/signup/signup.component.ts
@@ -81,7 +81,6 @@ export class SignupComponent implements OnInit, AfterViewInit, OnDestroy {
     private destroyed = false;
     private captchaRenderTimer: number | null = null;
     private usernameAvailabilityKey = '';
-    private usernameAvailabilityRequestId = 0;
     private readonly subs = new Subscription();
     private readonly customDomainPattern = /^(?=.{1,253}$)(?:[A-Za-z0-9](?:[A-Za-z0-9-]{0,61}[A-Za-z0-9])?\.)+[A-Za-z]{2,63}$/;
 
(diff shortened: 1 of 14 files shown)


src/app/signup/signup.component.ts
```diff
@@ -250,34 +252,26 @@ export class SignupComponent implements OnInit, AfterViewInit, OnDestroy {
...
-                            if (!this.runboxDomains.includes(this.runboxDomain)) {
-                                this.runboxDomain = this.runboxDomains[0];
-                            }
-                        }
-                        resolve();
-                    },
-                    error: () => resolve(),
-                });
-        });
+    private async loadRunboxDomains(): Promise<void> {
```
castaway: This does not return a promise (surprised it doesnt fail syntax checking) - should ideally use subscribe - see `profiles/profile.service.ts`

src/app/signup/signup.component.ts
```diff
@@ -250,34 +252,26 @@ export class SignupComponent implements OnInit, AfterViewInit, OnDestroy {
...
+                }
+            }
+        } catch {
+            // Keep safe defaults if the domain list cannot be loaded.
+        }
+    }
+
+    private async loadHCaptchaMetadata(): Promise<void> {
+        try {
+            this.hCaptchaSiteKey = await firstValueFrom(this.rmmapi.getSignupHCaptchaSiteKey());
```
castaway: See above re returning Promise and use of subscribe

castaway (changes requested):

src/app/signup/signup.component.ts
gtandersen: OK, will amend.

src/app/signup/signup.component.ts
gtandersen: Ditto.

castaway:
We should test:
* non-ascii characters in at least the first+last names
* on RMM signup vs Runbox7 signup, and check which fields get filled how

NB: I noticed that the RMM signup page figures out a timezone based on a country var in the page that the Template sets.. so we would need to replicate .. tho given the timezone is not amendable by the user, it could be fully done on the backend  (why isnt it amendable by the user?)

castaway (approved):
lgtm

PR #1870, opened 2026-06-10
fix(calendar): skip malformed ical events

### LINKED ISSUE #1484: Calendar: iCal events with errors prevent whole calendar loading
**Describe the bug**
Events created outside of Runbox7 might contain content that Runbox7 cannot parse. If those exist then the entire calendar does not load. Errors for example: `Object { message: "Missing parameter value in 'DESCRIPTION;ALTREP` are seen in the console. No errors are seen by the user.

**Expected behavior**
All parseable events should be loaded in the calendar, any that wont parse should produce an error message we can use to help the user fix the issue or report it to us.

keilogic:
## Summary

Closes #1484.

This makes calendar event loading tolerant of malformed iCalendar data. If one stored event fails while parsing or generating display occurrences, the service now reports that event through the existing `errorLog` path and keeps loading the remaining valid events.

The first commit adds the regression test by itself so it can be cherry-picked separately. The second commit contains the implementation.

## Tests

- `npx ng test --watch=false --browsers=FirefoxHeadless --include=src/app/calendar-app/calendar.service.spec.ts` - 10 SUCCESS
- `npx tsc -p tsconfig.json --noEmit`
- `git diff --check`
- Manual production build steps on Windows PowerShell:
  - `node src/build/pre-build.js`
  - `npx ng build runbox7 --configuration production --base-href /app/`
  - `node src/build/post-build.js`
- `npx ng test --watch=false --browsers=FirefoxHeadless` - 246 SUCCESS, 2 skipped
- `npm run lint` - exits 0 with the repository's existing warning baseline
- `npm run policy` - exits 0; it still prints existing historical commit-message warnings outside this branch

Note: the build pre-step regenerated `src/app/changelog/changes.ts`, but I did not include that generated file so this PR stays scoped to the calendar fix.

## AI Disclosure

I used OpenAI Codex while working on this PR. Codex assisted with editing `src/app/calendar-app/calendar.service.ts` and `src/app/calendar-app/calendar.service.spec.ts`; I reviewed the diff and ran the validation commands above locally.

### DIFF
--- a/src/app/calendar-app/calendar.service.spec.ts
+++ b/src/app/calendar-app/calendar.service.spec.ts
@@ -21,6 +21,7 @@ import { StorageService } from '../storage.service';
 import { RunboxWebmailAPI } from '../rmmapi/rbwebmail';
 import { CalendarService } from './calendar.service';
 import { RunboxCalendarEvent } from './runbox-calendar-event';
+import { HttpErrorResponse } from '@angular/common/http';
 import { of } from 'rxjs';
 import { take } from 'rxjs/operators';
 import moment from 'moment';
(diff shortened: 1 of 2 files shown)


src/app/calendar-app/calendar.service.spec.ts
```diff
@@ -226,6 +227,25 @@ END:VCALENDAR
         }));
     });
 
+    it('should skip malformed calendar events while loading valid ones', () => {
+        dav_events['test/broken'] = {
+            calendar: 'test',
+            id: 'test/broken',
+            ical: 'BEGIN:VCALENDAR\nBEGIN:VEVENT\nDESCRIPTION;ALTREP\nDTSTART:20210425T090000\nDTEND:20210425T100000\nSUMMARY:Broken\nEND:VEVENT\nEND:VCALENDAR\n'
```
castaway, line 234: Could you put the "broken" ical value not all on one long line (newlines are fine), and document what exactly is "broken" in the test data?

castaway (changes requested):

src/app/calendar-app/calendar.service.spec.ts
castaway: hi @keilogic could you respond to the requested changes?

PR #1989, opened 2026-06-15
fix(payments): Correct typo.

gtandersen:


castaway:
@gtandersen this one fails because it tries to load SignupPageComponent which it doesn't have - did that app.module.ts sneak in by accident?

PR #1999, opened 2026-08-03
style(login): Use more generic background and make alternative login links larger.

gtandersen:
Login page upgrade


src/styles.scss
```diff
@@ -421,6 +421,10 @@ body.signup-page app-rmm {
     justify-content: center;
     width: 100%;
     overflow: hidden;
+    background-image: url('https://runbox.com/wp-content/uploads/2022/07/sustainability2.jpg');
```
piratefinn, line 424: This is hotlinked from runbox.com WP. Realistic risk is the asset URL changing on a WP migration, then the login background is blank. Old setup used a bundled local asset. Either vendor it into `src/assets/`, or keep a `background-color` on `.loginSection` as a fallback (the white was moved off `.loginScreen` into `#loginArea`, so the section itself has nothing behind the image now).

src/app/login/login.component.html
```diff
@@ -62,7 +53,7 @@ <h2>Runbox Login</h2>
 	    </p>
 	    <div id="loginOptions">
 	      <mat-checkbox name="is_keep_logged" class="login-checkbox" title="Stay logged into Runbox until logging out manually" ngModel>Stay logged in</mat-checkbox>&nbsp;
-	      <mat-checkbox name="is_use_rmm6" id="useRmm6" class="login-checkbox" title="Use Runbox 6, the previous version of Runbox" ngModel>Use Runbox 6</mat-checkbox>
+	      <!--<mat-checkbox name="is_use_rmm6" id="useRmm6" class="login-checkbox" title="Use Runbox 6, the previous version of Runbox" ngModel>Use Runbox 6</mat-checkbox>-->
```
piratefinn, line 56: Commented out rather than removed, and `handleLoginResponse` in `login.component.ts:160` still reads `loginreq.is_use_rmm6` and redirects to `/mail`. Branch is unreachable now (2FA path passes `{}`, main path passes form value, and the checkbox's `ngModel` is gone so `is_use_rmm6` is never set). The `.side7` rule at `styles.scss:482` is orphaned too, all the `.side7` imgs were dropped from the template. Just delete the lot, commented HTML/CSS rots.

src/styles.scss
```diff
@@ -421,6 +421,10 @@ body.signup-page app-rmm {
...
+    background-size: 100%;
```
piratefinn, line 425: `100%` is width-only, image keeps aspect ratio against the width and gaps top/bottom on tall/narrow viewports. `cover` is the usual choice for full-bleed.

gtandersen: @piratefinn Thanks; these are fixed now.

castaway (approved):
lgtm (tested on rmmstage01)

PR #2000, opened 2026-08-23
feat(compose): move recipients between fields

### LINKED ISSUE #132: Support drag-and-drop of recipients
In Compose, allow recipients to be moved between recipient fields by dragging and dropping them.

Ref: https://community.runbox.com/t/contacts-to-field-moving-around/819

isaacsamual994-lang:
## Summary
- make existing To/CC/BCC recipient chips draggable
- preserve the source field during drag so dropping moves the recipient instead of copying it
- avoid duplicate recipients when dropping onto the same or target field

Closes #132.

## Testing
- Added regression tests before the implementation; both failed before the fix and pass afterward.
- Targeted recipient drag/drop tests: 2 passed.
- Full unit suite: 259 passed (2 skipped).
- `ng lint runbox7`: 0 errors (existing repository warnings remain).
- Production build: passed.

I also ran the repository CI wrapper locally. Its Cypress phase could not start Cypress 12.17.4 on the local darwin-arm64 environment (`Cypress.app: bad option: --no-sandbox`); the unit phase completed successfully before that environment-specific launcher failure.

### DIFF
--- a/src/app/compose/compose.component.html
+++ b/src/app/compose/compose.component.html
@@ -1,3 +1,4 @@
+
 <mat-card [ngClass]="{'draftPreview': !editing}" class="draft-card" [formGroup]="formGroup">
     <mat-card-actions class="draft-actions">
         <div class="draft-title">
@@ -51,6 +52,7 @@
                 [initialfocus]="model.to.length === 0"
                 placeholder="To"
                 [recipients]="model.to"
+                recipientField="to"
                 (updateRecipient)="onUpdateRecipient('to', $event)"
                 (drop)="recipientDropped($event, 'to')"
                 id="fieldTo"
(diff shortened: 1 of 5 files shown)


isaacsamual994-lang:
Implementation is ready for review. I re-verified the current PR branch locally: the targeted recipient drag/drop regression tests pass (2/2), `ng lint runbox7` completes with 0 errors (existing repository warnings only), and the production build passes. GitHub currently shows both workflows as awaiting maintainer approval and the PR as requiring one approving review. Could a maintainer please approve the workflows and review the PR when convenient? Also, could you confirm whether #132 is eligible for the Runbox 7 Feature Bounty Program and which reward tier applies?

PR #2001, opened 2026-08-27
fix(index): Increase index polling interval from 10 to 30 seconds.

gtandersen:
This change will decrease excessive requests to the servers while having a negligible effect to the user.


castaway (approved):
lgtm

